PR #26, opened 2015-01-12
Fix port typo in collectd.asciidoc

andis:
I believe there is a typo in the example config.


suyograo:
@andis thanks for this, but can you open this PR against https://github.com/logstash-plugins/logstash-codec-collectd. those docs get generated here

PR #45, opened 2015-02-05
Replace missing backticks

untergeek:


suyograo:
how about adding a link to http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/docs-bulk.html as reference

PR #56, opened 2015-02-18
Update pluginbody.asciidoc

zbuckholz:
github => 
does not work in many corporate networks because it's blocked, it needs to be changed to git => "http://github.com/......"


untergeek:
This needs to be pulled against the same file at https://github.com/elasticsearch/logstash/blob/master/docs/asciidoc/static/include/pluginbody.asciidoc

This repository is exclusively for our document publishing process.  We push changes from the primary logstash repo to this one.

PR #57, opened 2015-02-18
Update pluginbody.asciidoc

zbuckholz:


docs/static/include/pluginbody.asciidoc
```diff
@@ -1237,6 +1237,17 @@ and commit your changes.
 
 To publish version 0.1.0 of your new logstash gem:
 
+If you are using rvm as your ruby version manger, and rvm list shows jruby 1.7.19 as the current version you will need to
+set the following before running bundle exec rake publish_gem.
+export JRUBY_OPTS="--1.9"
+
+If you do not set --1.9 you will get the following errors.
+rake aborted!
+```
```
untergeek: Asciidoc is a bit different from Markdown.  Instead of using triple-backticks to get PRE-style quoting, you should encase the block in a line of dashes (or equals signs), rather like lines 1252-1257  just below:

> If you do not set --1.9 you will get the following errors.
>```
> ----------------------------------
> rake aborted!
> GemPublisher::CliFacade::Error: ERROR:  While executing gem ... (RuntimeError)
>     Can't find 'rubygems-update' in any repo. Check `gem source list`.
> ----------------------------------
>```

Don't worry about pruning extra lines (outside of the PRE block, anyway).  Asciidoc prefers them, and cleans up appropriately at render time.

untergeek:
I commented here, but can you instead do a pull against the regular Logstash repo? This doc technically resides at https://github.com/elasticsearch/logstash/blob/master/docs/asciidoc/static/include/pluginbody.asciidoc

This repo is exclusively for our publishing mechanism.  We push changes here after they've been merged in the regular Logstash repository.

Closing for now.

PR #72, opened 2015-04-08
Unix command tail -0F instead of -0a

DerekRoth:
Same issue as this one: https://github.com/elastic/logstash/issues/1927 and same fix.

### DIFF
--- a/docs/static/life-of-an-event.asciidoc
+++ b/docs/static/life-of-an-event.asciidoc
@@ -7,7 +7,7 @@ The Logstash event processing pipeline has three stages: inputs -> filters -> ou
 ==== Inputs
 You use inputs to get data into Logstash. Some of the more commonly-used inputs are:
 
-* *file*: reads from a file on the filesystem, much like the UNIX command `tail -0a`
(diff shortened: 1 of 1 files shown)


untergeek:
I apologize for the inconvenience, but this repository is the correct one for correcting docs.

I know, the Guide has those clever little "Edit" links, which take you here.  We're working on a way to get the metadata to point to the correct location, which in this case is https://github.com/elastic/logstash/blob/master/docs/asciidoc/static/life-of-an-event.asciidoc

As it states in the README for this repository, this is a generated repository and is not meant to be edited.

Again, please forgive us for the inconvenience this is and submit a pull request at https://github.com/elastic/logstash/blob/master/docs/asciidoc/static/life-of-an-event.asciidoc

Thanks for taking the time to catch and report and correct errors!

DerekRoth:
OK, the fact that there are "edit" buttons on the documentation page leading here was misleading. Anyway, i created another pull request on the correct repository.
Thanks

untergeek:
You are absolutely correct, and as I stated, we're actively looking for a way to make those point to the correct locations.  Thanks for being willing to move the PR.

Issue #73, opened 2015-04-10
Add logstash repository info from main branch


untergeek:
Merge in new doc from https://github.com/elastic/logstash/pull/2981 and add link to index.

PR #74, opened 2015-04-10
This addresses #73

untergeek:
Add repository information.

This also prevents the html_docs folder from being accidentally
merged by a `commit -a` by way of the `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+docs/html_docs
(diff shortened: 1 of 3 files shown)


untergeek:
We should consider what we want this to look like:

screenshot 2015-04-10 11 39 18

Do we want the bare word, "Repositories," or do we want something like "Package Repositories" or something like that?

untergeek:
The results: 

screenshot 2015-04-10 11 38 26

untergeek:
Update:

screenshot 2015-04-10 11 55 28

untergeek:
And:

screenshot 2015-04-10 11 56 09

PR #99, opened 2015-07-03
Updates index file to include text expansion macros for links.

palecur:
The current index file includes deploying.asciidoc, but not the added macros that make the inline links work.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,10 +1,14 @@
 [[logstash-reference]]
 = Reference
 
-:branch:          1.5
-:logstash_version: 1.5.0
+:branch:                1.5
+:logstash_version:      1.5.0
 :elasticsearch_version: 1.5.1
-:jdk:     1.7.0_60
+:jdk:                   1.7.0_60
+:guide:                 https://www.elastic.co/guide/en/elasticsearch/guide/current/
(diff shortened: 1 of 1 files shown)


docs/index.asciidoc
```diff
@@ -1,10 +1,14 @@
 [[logstash-reference]]
 = Reference
 
-:branch:          1.5
-:logstash_version: 1.5.0
+:branch:                1.5
+:logstash_version:      1.5.0
```
ph: should it be 1.5.2 ?

docs/index.asciidoc
```diff
@@ -1,10 +1,14 @@
...
 :elasticsearch_version: 1.5.1
```
ph: Probably worth changing this for 1.6.0 ?

docs/index.asciidoc
palecur: Since @suyograo didn't change it when he updated this index doc earlier today, I figured he had a reason and decided to leave it as is.

docs/index.asciidoc
palecur: Probably -- checking with @suyograo now.

palecur:
While the version number discussion is important, let's fix them later -- right now this incorrect index file is resulting in links not rendering and lines missing from the doc on the site.

PR #109, opened 2015-07-31
Modifies index to support new getting started guide

palecur:
Adds index file entry to include new asciidoc file.


jordansissel:
LGTM as far as I can tell :)

PR #137, opened 2015-11-11
Removing LSF from roadmap

acchen97:
Removing LSF refs from public roadmap


jordansissel:
Change LGTM; Let's make sure this change lands in the logstash repo (same file to modify) also.

PR #144, opened 2015-12-12
Adds examples for presence/absence tests for fields

palecur:


jordansissel:
LGTM

jordansissel:
(One minor quibble - that the commit message doesn't mention some of the changes, such as mentioning `@metadata` only occurs in 1.5+, or similar)

palecur:
There's an 'added in 1.5' floater in the title of thes ection, is that insufficient?

jordansissel:
no no I mean that your commit message doesn't declare that you intended to include that change - sorry, I am being confusing.

I read commit messages as an author's description of their intent with respect to a change, so if I see a change in the text not described in the commit message, I wonder - does that make sense?

For this PR, I am not worried, but future PRs we can improve the commit messages. ;P

PR #147, opened 2016-01-19
Fixed cloudwatch section headers

leedm777:
The asciidoc marker for headers is `=`, not `#`

### DIFF
--- a/docs/plugins/outputs/cloudwatch.asciidoc
+++ b/docs/plugins/outputs/cloudwatch.asciidoc
@@ -5,7 +5,8 @@
 
 This output lets you aggregate and send metric data to AWS CloudWatch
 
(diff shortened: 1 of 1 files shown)


untergeek:
@leedm777 This repository is not meant to be edited directly.  All documentation exists in either https://github.com/elastic/logstash, or in its respective repository in https://github.com/logstash-plugins.  We build the docs, then push them to this repository for publication.

As such, the place the code changes need to go is at: https://github.com/logstash-plugins/logstash-output-cloudwatch/blob/master/lib/logstash/outputs/cloudwatch.rb

leedm777:
Moved to logstash-plugins/logstash-output-cloudwatch#7

PR #234, opened 2016-07-08
Change bin/plugin to bin/logstash-plugin

dedemorton:


dedemorton:
Made changes directly to the logstash-docs repo to replace bin/plugin with bin/logstash-plugin in the published doc. This fixes the doc renaming issue listed in the doc meta issue here: https://github.com/elastic/logstash/issues/5148.

I've also made this change in the 2.3 and master branches.

@suyograo It looks like you've already updated https://github.com/elastic/logstash/blob/5.0/docs/plugin-doc.asciidoc.erb. This means that the plugin doc should be correct the next time you generate the Logstash plugin docs, right?

PR #241, opened 2016-07-13
Add kafka compatibility matrix

dedemorton:


suyograo:
@dedemorton can we add the following sentence after/before the table.

" Please remember that it is good advice to upgrade brokers before consumers/producers since brokers target backwards compatibility. For example, the 0.9 broker will work with both the 0.8 consumer and 0.9 consumer APIs but not the other way around."

suyograo:
LGTM

dedemorton:
This PR should be complete now. I'm going to make the changes in the ruby code in the logstash repo and then publish the changes.

PR #269, opened 2016-08-23
Update docs for S3 input to reflect the realities of the code.

joestump:
The [docblock](https://github.com/logstash-plugins/logstash-input-s3/blob/master/lib/logstash/inputs/s3.rb#L27-L31) on `sincedb_path` suggests the default isn't `nil`. [Here's where the default is set](https://github.com/logstash-plugins/logstash-input-s3/blob/master/lib/logstash/inputs/s3.rb#L273-L275) to `$HOME/.sincedb_{hash}`.

### DIFF
--- a/docs/plugins/inputs/s3.asciidoc
+++ b/docs/plugins/inputs/s3.asciidoc
@@ -233,7 +233,7 @@ The AWS Session token for temporary credentials
 ===== `sincedb_path` 
 
   * Value type is <<string,string>>
-  * Default value is `nil`
+  * Default value is `$HOME/.sincedb_{MD5(bucket + prefix)}`
 
 Where to write the since database (keeps track of the date
 the last handled file was added to S3). The default will write


dedemorton:
@joestump Thanks for your PR. However, please note that the files in this repo are generated from the Ruby source code in https://github.com/logstash-plugins/logstash-input-s3 (we don't accept PRs in the logstash-docs repo for doc changes). Any changes that we make to the asciidoc files in logstash-docs will be lost when the plugin docs are generated again.

Note that the defaults shown in the docs come from the defaults specified in the code here:
https://github.com/logstash-plugins/logstash-input-s3/blob/master/lib/logstash/inputs/s3.rb (see [Line 31](https://github.com/logstash-plugins/logstash-input-s3/blob/master/lib/logstash/inputs/s3.rb#L31) for the sincedb_path default). 

Any changes to the generated docs need to be made to the source code in the plugin repository. Let me know if you have any questions.

Closing this PR.

PR #296, opened 2016-09-22
sort the list alphabetically

ph:


ph:
@suyograo does the `edit_url` section need to be in the same order as the table or it doesn't matter.

PR #297, opened 2016-09-23
Sort the list of plugin alphabetically

ph:


dedemorton:
The content here is a bit stale. These changes will be in the PR that I'm adding for community vs elastic maintained plugins, so I'm closing this now.

PR #342, opened 2016-12-03
Consolidate plugin list

acchen97:
Consolidating plugin list, to merge at 5.1 launch.

### DIFF
--- a/docs/plugins/codecs.asciidoc
+++ b/docs/plugins/codecs.asciidoc
@@ -4,55 +4,33 @@
 A codec plugin changes the data representation of an event. Codecs are essentially stream filters that can operate as part 
 of an input or output.
 
-The following codec plugins are available:
-
-* <<elastic-supported-codec-plugins,Elastic supported plugins>>
-* <<community-supported-codec-plugins,Community supported plugins>>
-
-[float]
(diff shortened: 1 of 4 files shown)


acchen97:
@dedemorton @suyograo should be a quick review, will merge alongside 5.1 release.

dedemorton:
Looks Groovy To Me (the files build cleanly). Thanks for not making me do this. :-)

dedemorton:
@acchen97 So just to clarify, you want us to wait until 5.1 to merge this, right? And then we should merge it into 5.0, 5.1, 5.x, and master, right?

acchen97:
@dedemorton yep will make sure to do so, thanks for the review.

suyograo:
LGTM

jordansissel:
Wait. Why was this done on Logstash-docs? Shouldn't it be in the Logstash
repo under docs?  Reason being is the next docs sync job will probably
revert this change.

dedemorton:
@jordansissel There are a few files in the plugin docs that are still static and need to be updated in logstash-docs (the overview files for each plugin type). We do plan to automate this eventually, but right now, the single truth of source for this file is in logstash-docs.

PR #352, opened 2017-02-05
fix yaml link

oxnz:
http//yaml.org/ -> http://yaml.org/


jordansissel:
Thanks for this fix. Can you love this PR into the logstash repo? This
repository (logstash-docs) is not for humans and only exists to feed the
Elastic docs system

dedemorton:
This has been fixed in a separate PR, so I'm closing it.

PR #355, opened 2017-03-28
Add plugin level required to support sending content-type headers

dedemorton:
@acchen97 Just an FYI that I'm adding these notes to the LS 2.4 docs.


acchen97:
Awesome thanks!

PR #384, opened 2017-06-21
[DOCS] Copy index from logstash repo

lcawl:
This PR copies the updated index.asciidoc, index-shared1.asciidoc, and index-shared2.asciidoc from the logstash repo, since @dedemorton noted that they aren't copied automatically.

The same will need to be done for the 2.3, 2.2, 2.1, 2.0, and 1.5 versions of these files.


dedemorton:
LGTM.

Nothing here that the build now pulls these files from the logstash repo. We're just adding them to logstash-docs to make sure the "static" files in the two repos are in sync.

@lcawl As I mentioned, I'll take care of copying over the files from 1.5->2.3. Thanks!!

dedemorton:
@lcawl Actually since the shared files are not included in this PR, I'll take care of all branches and close this. Thanks again!

PR #386, opened 2017-06-24
Draft of plugin doc fixes 5.5

dedemorton:
DON'T MERGE THIS.

I'm jsut creating a pull request to document the changes required to the 5.5 docs. 

@ph @suyograo Here's the stuff that caused problems in 5.5. Pretty similar to master, as expected. Not sure why useragent didn't updates for the ID change that you made. 

I notice that the plugin name change got into the script, but not the abbreviated title. I think that's OK, but if it's possible to get the abbreviated title, that would be cool, too. I'm just happy that it's building at this point. 

Here's a [build of the plugin docs](https://logstashplugins.firebaseapp.com/). It doesn't include the new stuff, like dead letter queue, because I need to add it to the index. Wanted to test the existing ones first.

### DIFF
--- a/docs/plugins/codecs.asciidoc
+++ b/docs/plugins/codecs.asciidoc
@@ -37,53 +37,78 @@ The following codec plugins are available below. For a list of Elastic supported
 
 :edit_url: https://github.com/logstash-plugins/logstash-codec-avro/edit/master/docs/index.asciidoc
 include::codecs/avro.asciidoc[]
+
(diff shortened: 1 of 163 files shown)


ph:
@dedemorton These files https://github.com/elastic/logstash-docs/pull/386/files#diff-d28b7cfef131ca97ec437d3060c77d5a are generated manually right? I don't think we have touched them?

PR #387, opened 2017-06-28
Update plugin docs for 5.5 - patched versions

dedemorton:
Here's a "patched" version of the plugin docs based on the docs that were generated two days ago. I've verified that all the version info matches the installed plugins for 5.5. I propose that we use this version of the plugin content if we cannot regenerate the docs in time for the 5.5 release. We can quickly release a new version of the doc, but I think this gets us pretty close. If the docs reflect incorrect details about the community-maintained plugins, I think that's OK...for now.

I've created lots of commits intentionally so I can cherry-pick any relevant changes if we do get the docgen to work successfully tomorrow.

TO DO: If we use these files, I'll make all the topic titles consistent. 

There were still quite a few files that were old, so here's a description of what I did with content that wasn't quite right:

* Updated asciidoc in the following plugins:

logstash-input-eventlog -  content was very old. I was able to install v4.1.1 on 5.5, so I updated the docs to reflect that version.
logstash-filter-truncate - content was old - I was able to install v1.0.2 on 5.5, so I updated the docs to reflect that version.
logstash-input-gelf - changed version - docs showed version as 3.0.5, but should be v3.0.4, so I changed the version 
logstash-input-xmpp - changed version - docs showed version as 3.1.4, but should be 3.1.3, so I changed the version

* Old docs that still  had the old community maintainer note. I handled these differently based on the install issues that I encountered:

When install resulted in errors about incompatible versions, I changed the community maintainer note in these plugins to indicate that the plugin could not be installed on the current version of logstash. If a note didn't exist, I added one. Was this the right thing to do?

logstash-input-gemfire
logstash-input-rackspace
logstash-input-websocket
logstash-input-zenoss
logstash-input-zeromq
logstash-filter-cipher
logstash-filter-collate
logstash-filter-metaevent
logstash-filter-punct
logstash-filter-yaml
logstash-filter-zeromq
logstash-codec-oldlogstashjson
logstash-output-rackspace
logstash-output-zeromq

When install failed with "error occurred while installing net-ssh",  I left the community-maintainer note as it was. What should I do here?

logstash-input-heroku
logstash-codec-cloudtrail
logstash-codec-compress_spooler - 
logstash-output-gelf
logstash-output-google_cloud_storage
logstash-output-jira
logstash-output-newrelic

- The following plugin installed, but it was an old version, so I  left the community-maintainer note. What should I do here?

drupal_dblog input - old formatting - didn't update because latest installable version was 2.0.3 - March 18, 2016 java (905 KB) 

- The following plugin does not exist, so I need someone to confirm that I should delete it from the index and overview table:

logstash-codec-s3_plain  

- This file exists, but it's not in the TOC. Should I delete it:

logstash-codec-spool

### DIFF
--- a/docs/plugins/codecs.asciidoc
+++ b/docs/plugins/codecs.asciidoc
@@ -24,7 +24,7 @@ The following codec plugins are available below. For a list of Elastic supported
 | <<plugins-codecs-json,json>> | Reads JSON formatted content, creating one event per element in a JSON array | https://github.com/logstash-plugins/logstash-codec-json[logstash-codec-json]
 | <<plugins-codecs-json_lines,json_lines>> | Reads newline-delimited JSON | https://github.com/logstash-plugins/logstash-codec-json_lines[logstash-codec-json_lines]
 | <<plugins-codecs-line,line>> | Reads line-oriented text data | https://github.com/logstash-plugins/logstash-codec-line[logstash-codec-line]
-| <<plugins-codecs-msgpack,msgpack>> |  | https://github.com/logstash-plugins/logstash-codec-msgpack[logstash-codec-msgpack]
+| <<plugins-codecs-msgpack,msgpack>> | Reads MessagePack encoded content | https://github.com/logstash-plugins/logstash-codec-msgpack[logstash-codec-msgpack]
 | <<plugins-codecs-multiline,multiline>> | Merges multiline messages into a single event | https://github.com/logstash-plugins/logstash-codec-multiline[logstash-codec-multiline]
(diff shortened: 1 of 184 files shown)


dedemorton:
Closing because https://github.com/elastic/logstash-docs/pull/388 merges all the changes here (plus the changes in plugin-test/5.5 into the 5.5

dedemorton:
@jordansissel Ignore my previous comment. I'm decided that it actually made more sense to merge this into the test branch so it matches what we have in 5.5.

suyograo:
> When install resulted in errors about incompatible versions, I changed the community maintainer note in these plugins to indicate that the plugin could not be installed on the current version of logstash. If a note didn't exist, I added one. Was this the right thing to do?

Good to add this banner. We should change the wording to include:

This plugin cannot be installed on the current version of Logstash. We are working on resolving this.

suyograo:
Thanks for hand editing these @dedemorton. <3

PR #405, opened 2017-07-13
Add note about updating DLQ plugin

dedemorton:
@acchen97 @suyograo I thought we might want to add this to be consistent with what we did with the dissect filter. Feel free to merge if you want to add this.

This note is in the read-only repo, not the source repos, so it will get overwritten when we regenerate the plugin docs.


acchen97:
Good idea, thanks @dedemorton LGTM.

PR #407, opened 2017-08-18
Generate docs from 5.6 branch / #8031

jordansissel:


dedemorton:
@jordansissel I was able to get a clean doc build (no errors...yay). I noticed that some of the plugins didn't get updated. For the previous 5.x release, we ran into the same problem and added a banner to the documentation that says: "This plugin cannot be installed on the current version of Logstash. We are working on resolving this problem." The banner was meant to be a stop-gap while we figured out what to do. @suyograo  opened an issue, but looks like there has been no movement: https://github.com/elastic/logstash/issues/7576

I'm wondering what we should do with this content because the banner suggests we are actively fixing the problem. 

The old topics sort of stand out as different in the TOC because they don't use the abbreviated titles that I added to make the TOC easier to scan. Looks a little weird, but it's not a deal breaker.

image

jordansissel:
> The banner was meant to be a stop-gap while we figured 

I propose we just remove the plugins from the docs until they are fixed.

jordansissel:
@dedemorton thoughts?

We could remove them today and add the plugin docs back in once the plugins themselves are working.

jordansissel:
> The old topics sort of stand out as different in the TOC 

We can fix these, but I suspect they will require plugin version bumps unless we make it possible to publish docs without plugin releases.

Ideas?

dedemorton:
>We could remove them today and add the plugin docs back in once the plugins themselves are working.

@jordansissel If possible, I'd like @suyograo to weigh in here because I remember he didn't want to remove the docs. I'd prefer removing the docs if the plugins are old (and maybe lower quality) and we'll end up removing them anyhow. I would like to make a pass at the plugins that I flagged as uninstallable for 5.5, but don't show up on your list, to figure out if there is still an install problem with those specific plugins.

IMO we should go ahead and publish the plugins that you built for 5.6 (even though they are not perfect). We can always remove topics early next week. It's annoying that the TOC looks weird, but at least users will be able to see the updated content for 5.6, and we'll have a couple of days to sort out what to do with the broken plugins. WDYT?

> We can fix these, but I suspect they will require plugin version bumps unless we make it possible to publish docs without plugin releases.

I don't think bumping versions will help in this case because the titles live in plugins that aren't building. :-)

PR #428, opened 2017-11-14
Remove docs for anonymize filter

dedemorton:
@acchen97 I've removed the anonymize filter docs as requested. This is correct, no? Let me know if there are any other plugin docs that need to be removed. At this point, I believe we are building docs for everything that is installable on 6.0.0.

@jsvd Just a heads-up that I've manually altered this file. I'm not sure how the script you're working on (to automate the creation of these overview files) will handle plugin docs for plugins that we no longer want to document but are still installable.


acchen97:
Yes, correct, starting in 6.0.

dedemorton:
cherry-picked into master + 6.x, so closing now.

PR #432, opened 2017-11-30
update docs for 6.1.0

jsvd:


docs/plugins/filters/ruby.asciidoc
```diff
@@ -32,15 +38,100 @@ For example, to cancel 90% of events, you can do this:
...
+def filter(event)
+	if rand >= @drop_percentage
+		return [event]
+	else
+		return [] # return empty array to cancel event
+	end
+end
+----
+
+====== Testing the ruby script
```
dedemorton: Level 6 headings are not supported by the doc build. This renders as text. Change this to ===== Testing...

docs/plugins/filters/ruby.asciidoc
```diff
@@ -32,15 +38,100 @@ For example, to cancel 90% of events, you can do this:
...
+
+  expect("drops the event") do |events|
+    events.size == 0
+  end
+end
+----
+
+We can now test that the ruby script we're using is implemented correctly:
+
+[source]
```
dedemorton: This results in a doc build error because it should be [source,shell]

docs/plugins/filters/ruby.asciidoc
```diff
@@ -50,8 +141,11 @@ This plugin supports the following configuration options plus the <<plugins-{typ
 [cols="<,<,<",options="header",]
 |=======================================================================
 |Setting |Input type|Required
-| <<plugins-{type}s-{plugin}-code>> |<<string,string>>|Yes
+| <<plugins-{type}s-{plugin}-code>> |<<string,string>>|No
 | <<plugins-{type}s-{plugin}-init>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-path>> |<<string,string>>|No
+| <<plugins-{type}s-{plugin}-script_params>> |<<hash,hash>>,{}|No
+| <<plugins-{type}s-{plugin}-tag_on_exception>> |<<string,string>>,_rubyexception|No
```
dedemorton: This ID causes a doc build error because the description of the tag_on_exception config option doesn't exit.

dedemorton: Maybe this should be tag_on_failure?

jsvd:
@dedemorton I fixed these issues in https://github.com/logstash-plugins/logstash-filter-ruby/pull/38, we can either wait for the patch release of the ruby filter or edit these docs , wdyt?

dedemorton:
@jsvd LGTM but we need to add the changes from https://github.com/logstash-plugins/logstash-filter-ruby/pull/38 added. Let's edit the ruby.asciidoc file rather than waiting for the patch because we need the fixes from https://github.com/logstash-plugins/logstash-filter-ruby/pull/38 in order to build the doc without errors. Andrew pinged me just this morning to make sure that these docs will be available for 6.1.

jsvd:
@dedemorton I've updated the ruby filter docs with the fixed version, this should be good to merged.

PR #433, opened 2017-12-01
Create prototype of versioned plugin docs

dedemorton:
This PR contains a prototype for the Versioned Plugin Reference. The goal is to create a Versioned Plugin Reference that contains multiple versions of the documentation for each plugin.

To see this content rendered, pull the files to your system and run:

````
../docs/build_docs.pl --doc ../logstash-docs/docs/versioned-plugins/index.asciidoc --chunk=1 -open
````

(adjust the paths for your environment)

I've copied the include folder from the logstash repo and modified the files to work in a sandbox. The final version will used the shared files in the logstash repo.

The actual plugin content isn't really versioned (just copies of the same file) so the dates are the same.

Ping me if you want a link to the detailed document outlining the changes that are required to make this all work.

### DIFF
--- a/docs/versioned-plugins/codecs.asciidoc
+++ b/docs/versioned-plugins/codecs.asciidoc
@@ -1,0 +1,6 @@
+:type: codec
+:type_uc: Codec
+
+include::include/plugin-intro.asciidoc[]
+
+include::codecs/rubydebug-index.asciidoc[]
--- a/docs/versioned-plugins/codecs/rubydebug-index.asciidoc
+++ b/docs/versioned-plugins/codecs/rubydebug-index.asciidoc
@@ -1,0 +1,14 @@
+:plugin: rubydebug
+:type: codec
+
+include::{include_path}/version-list-intro.asciidoc[]
+
+|=======================================================================
+| Version | Release Date 
+| <<v3.0.4-plugins-codecs-rubydebug,v3.0.4>> | Aug 21, 2017 (latest)
+| <<v3.0.3-plugins-codecs-rubydebug,v3.0.3>> | Aug 21, 2017
+|=======================================================================
+
+
+include::rubydebug-v3.0.4.asciidoc[]
+include::rubydebug-v3.0.3.asciidoc[]
--- a/docs/versioned-plugins/codecs/rubydebug-v3.0.3.asciidoc
+++ b/docs/versioned-plugins/codecs/rubydebug-v3.0.3.asciidoc
@@ -1,0 +1,47 @@
+:plugin: rubydebug
+:type: codec
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: v3.0.3
+:release_date: 2017-08-21
+:changelog_url: https://github.com/logstash-plugins/logstash-codec-rubydebug/blob/v3.0.4/CHANGELOG.md
+:include_path: ../include
+:logstash-ref: http://www.elastic.co/guide/en/logstash/6.0
+///////////////////////////////////////////
+END - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+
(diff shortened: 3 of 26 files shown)

### CODE REVIEW

File: docs/versioned-plugins/codecs.asciidoc [hunk 1]
```diff
@@ -0,0 +1,6 @@
+:type: codec
+:type_uc: Codec
+
+include::include/plugin-intro.asciidoc[]
+
+include::codecs/rubydebug-index.asciidoc[]
```
jsvd: this is really easy to template and expand with the list of codecs retrieved from the logstash-plugins github org

docs/versioned-plugins/codecs/rubydebug-index.asciidoc
```diff
@@ -0,0 +1,14 @@
+:plugin: rubydebug
+:type: codec
+
+include::{include_path}/version-list-intro.asciidoc[]
+
+|=======================================================================
+| Version | Release Date 
+| <<v3.0.4-plugins-codecs-rubydebug,v3.0.4>> | Aug 21, 2017 (latest)
+| <<v3.0.3-plugins-codecs-rubydebug,v3.0.3>> | Aug 21, 2017
```
jsvd: this also seems easy to template, I wonder if this release date should be the one corresponding to the github tag or the rubygems release date (they'll likely be the same, but not necessarily).

docs/versioned-plugins/codecs/rubydebug-v3.0.3.asciidoc
```diff
@@ -0,0 +1,47 @@
...
+:type: codec
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: v3.0.3
+:release_date: 2017-08-21
+:changelog_url: https://github.com/logstash-plugins/logstash-codec-rubydebug/blob/v3.0.4/CHANGELOG.md
+:include_path: ../include
+:logstash-ref: http://www.elastic.co/guide/en/logstash/6.0
```
jsvd: Is `logstash-ref` going to always point to the latest release? if so, this also needs to be templated.

dedemorton: that's tbd. I've suggested adding the attribute to each doc file so that we aren't locked into having logstash-ref be the same value for every plugin.

jordansissel:
image

This is delightful. Thank you for this POC, @dedemorton!

docs/versioned-plugins/include/version-list-intro.asciidoc
```diff
@@ -0,0 +1,13 @@
...
+== Versioned {plugin} {type} plugin docs
+++++
+<titleabbrev>{plugin}</titleabbrev>
+++++
+
+This page lists all available versions of the documentation for this plugin. 
+To see which version of the plugin you have installed, run `bin/logstash-plugin
+list --verbose`. 
+
+NOTE: Versioned plugin documentation is not available for plugins released prior
```
jsvd: I think the actual restriction here is we can only do this easily for plugin versions that have the separated index.asciidoc (instead of the inline documentation), no? That said, I agree we can put a "starting marker" in the Logstash 6.0 release

dedemorton: Yup... 6.0 is just a placeholder for whatever we want to say based on how far back we want to go. We were still making manual hacks to the generated asciidoc during the 5.x releases, so we probably don't want to go back too far or we'll have a bunch of broken asciidoc stuff to fix.  We could be a bit wishy washy and say something like, "Versioned plugin documentation was added to the Elastic docs during the Logstash 6.0 release. Therefore the documentation for plugins released prior to 6.0 might not be available here."

jordansissel:
Overall I am greatly in favor of this proposal. 👍

dedemorton:
Closing because we've moved on to a newer version of the prototype.

PR #444, opened 2018-01-11
Versioned plugin docs

jsvd:
this is a demo of the generated docs for versions of all plugins that have a `docs/index.asciidoc`


docs/versioned-plugins/outputs/elasticsearch-v7.3.2.asciidoc
```diff
@@ -123,7 +123,7 @@ This plugin supports the following configuration options plus the <<{version}-pl
 | <<{version}-plugins-{type}s-{plugin}-pipeline>> |{logstash-ref}/configuration-file-structure.html#string[string]|No
 | <<{version}-plugins-{type}s-{plugin}-pool_max>> |{logstash-ref}/configuration-file-structure.html#number[number]|No
 | <<{version}-plugins-{type}s-{plugin}-pool_max_per_route>> |{logstash-ref}/configuration-file-structure.html#number[number]|No
-| <<{version}-plugins-{type}s-{plugin}-proxy>> |{logstash-ref}/configuration-file-structure.html#uri[uri]|No
```
dedemorton: You missed a type here.

dedemorton:
@jsvd We should talk about how we are going to iterate over these files. I want to avoid having to manually fix stuff over and over again. Once we get a set of files building without errors, can we say that they are done and iterate over a different set of files for the next round of fixes? 

Is that possible? I'd like to say that those versions of the plugin docs are done. After we've gone through all the plugins and tweaked the script, I'll make any changes required to the plugin repos (mostly for external links, I suspect), and then we'll only run into errors when someone checks in asciidoc that breaks something. If we're only building the new versions, then we should never have to generate the older ones again, right? WDYT? IT might be worth meeting to discuss this.

jsvd:
closing as we're moving to other rounds of plugin generation

PR #445, opened 2018-01-13
[Work in progrss] Versioned plugin doc fixes

dedemorton:


docs/versioned-plugins/codecs/cef-v4.1.3.asciidoc
```diff
@@ -33,22 +33,22 @@ produce an event with the payload as the 'message' field and a '_cefparsefailure
...
 |Setting |Input type|Required
-| <<{version}-plugins-{type}s-{plugin}-delimiter>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-fields>> |<<array,array>>|No
-| <<{version}-plugins-{type}s-{plugin}-name>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-product>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-severity>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-signature>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-vendor>> |<<string,string>>|No
-| <<{version}-plugins-{type}s-{plugin}-version>> |<<string,string>>|No
+| <<{version}-plugins-{type}s-{plugin}-delimiter>> |{logstash-ref}/configuration-file-structure.html#string[string]|No
```
dedemorton: The script needs to replace these external links to point to the Logstash Reference.

docs/versioned-plugins/codecs/cef-v4.1.3.asciidoc
```diff
@@ -12,7 +12,7 @@ START - GENERATED VARIABLES, DO NOT EDIT!
 END - GENERATED VARIABLES, DO NOT EDIT!
 ///////////////////////////////////////////
 
-[id="plugins-{type}-{plugin}"]
+[id="{version}-plugins-{type}s-{plugin}"]
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin. We can fix it with a simple search/replace in the old files.

docs/versioned-plugins/filters/grok-v3.4.1.asciidoc
```diff
@@ -7,12 +7,12 @@ START - GENERATED VARIABLES, DO NOT EDIT!
 :version: v3.4.1
 :release_date: 2017-05-10
 :changelog_url: https://github.com/logstash-plugins/logstash-filter-grok/blob/v3.4.1/CHANGELOG.md
-:include_path: ../../../logstash/docs/include
+:include_path: ../../../../logstash/docs/include
```
dedemorton: The path is wrong here. This problem is caused by bad asciidoc in older versions of the plugin. We can fix it with a simple search/replace in the old files.

docs/versioned-plugins/filters/grok-v3.4.1.asciidoc
```diff
@@ -160,33 +160,33 @@ filter. This newly defined patterns in `pattern_definitions` will not be availab
 [id="{version}-plugins-{type}s-{plugin}-options"]
 ==== Grok Filter Configuration Options
 
-This plugin supports the following configuration options plus the <<plugins-{type}s-common-options>> described later.
+This plugin supports the following configuration options plus the <<{version}-plugins-{type}s-{plugin}-common-options>> described later.
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin. We can fix it with a simple search/replace in the old files.

docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc
```diff
@@ -124,7 +124,7 @@ end
 
 We can now test that the ruby script we're using is implemented correctly:
 
-[source]
+[source,shell]
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin. We can fix it with a simple search/replace in the old files.

docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc
```diff
@@ -182,11 +182,20 @@ The path of the ruby script file that implements the `filter` method.
...
 ===== `script_params`
 
-  * Value type is <<hash,hash>>
+  * Value type is {logstash-ref}/configuration-file-structure.html#hash[hash]
   * Default value is `{}`
 
 A key/value hash with parameters that are passed to the register method
 of your ruby script file defined in `path`.
 
+[id="{version}-plugins-{type}s-{plugin}-tag_on_exception"]
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin. We can fix it by adding the missing content to all the relevant files.

docs/versioned-plugins/filters/ruby-v3.1.1.asciidoc
```diff
@@ -86,7 +86,7 @@ def filter(event)
 end
 ----
 
-====== Testing the ruby script
+===== Testing the ruby script
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin. We can fix it with a simple search/replace in the old files.

docs/versioned-plugins/index.asciidoc
```diff
@@ -11,10 +11,10 @@ include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 [[logstash-plugin-reference]]
 = Versioned Plugin Reference
 
-include::inputs.asciidoc[]
+include::inputs-index.asciidoc[]
```
dedemorton: The filenames used in the index file are not correct. We need to fix this in the index.asciidoc file.

docs/versioned-plugins/inputs/beats-v4.0.1.asciidoc
```diff
@@ -46,57 +46,57 @@ output {
 
 NOTE: The Beats shipper automatically sets the `type` field on the event.
 You cannot override this setting in the Logstash config. If you specify
-a setting for the <<plugins-inputs-beats-type,`type`>> config option in
+a setting for the <<{version}-plugins-inputs-beats-type,`type`>> config option in
```
dedemorton: This is an internal link, so it needs to include {version}. We'll need to fix the script to add {version} to all internal links. However, it might make more sense to do this *after* we've converted the format of all external links in asciidoc source in the various repos. It's going to be a bit of cleanup work, and probably manual.

docs/versioned-plugins/inputs/beats-v4.0.1.asciidoc
```diff
@@ -46,57 +46,57 @@ output {
...
 Logstash, it is ignored.
 
 IMPORTANT: If you are shipping events that span multiple lines, you need to
 use the https://www.elastic.co/guide/en/beats/filebeat/current/multiline-examples.html[configuration options available in Filebeat] to handle multiline events
 before sending the event data to Logstash. You cannot use the
-<<plugins-codecs-multiline>> codec to handle multiline events. Doing so will
+{logstash-ref}/plugins-codecs-multiline.html[Multiline codec plugin] codec to handle multiline events. Doing so will
```
dedemorton: This is an external link, so we need to convert it to use the format for external links. We'll need to fix this in the asciidoc in the plugin repo and then use search/replace to fix it in older docs for the plugin.

docs/versioned-plugins/inputs/beats-v4.0.1.asciidoc
```diff
@@ -229,12 +229,12 @@ The maximum TLS version allowed for the encrypted connections. The value must be
...
-  * Value type is <<number,number>>
+  * Value type is {logstash-ref}/configuration-file-structure.html#number[number]
   * Default value is `1`
 
 The minimum TLS version allowed for the encrypted connections. The value must be one of the following:
 1.0 for TLS 1.0, 1.1 for TLS 1.1, 1.2 for TLS 1.2
 
 
-
+[id="{version}-plugins-{type}s-{plugin}-common-options"]
```
dedemorton: This problem is caused by bad asciidoc in older versions of the plugin where this line is missing. We can fix it by updating the files where this is missing.

docs/versioned-plugins/outputs/email-v4.0.4.asciidoc
```diff
@@ -12,14 +12,16 @@ START - GENERATED VARIABLES, DO NOT EDIT!
...
-[id="plugins-{type}-{plugin}"]
+[id="{version}-plugins-{type}s-{plugin}"]
 
 === Email output plugin {version}
 
 include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
+Sends email when an output is received. Alternatively, you may include or
```
dedemorton: This description is missing from the older plugins. It must be added, or the build fails.

jsvd:
@dedemorton I have updated https://github.com/elastic/logstash-docs/pull/444 with a few fixes for the things pointed out in this PR.

Things missing:
- detecting missing description contents and id tags
- converting external links

One question, I see the links `<<{version}-plugins-{type}s-{plugin}-common-options>>`

Should the common options link to a specific plugin (and specific plugin version)?

dedemorton:
@jsvd Yes. The version info is required in the links because section with common options gets pulled into every plugin file through include statements. If you don't include the version info, then the IDs for those sections will not be unique in each file, and the doc build will fail with errors.

You'll notice in my latest commit that I point to the include folder in the logstash repo rather than the one under logstash-docs/docs/versioned-plugins. I think we probably want one source of truth for the common files, so I updated the common files in the logstash repo so that they work with both the versioned plugin doc build and the Logstash Reference build: elastic/logstash#8937. But as I mention in my comments earlier, we need to decide how we will handle the common files over time. I want to maintain one source so we don't have duplicated content, but we might want to take the files, copy them to logstash-docs, and version them somehow (see the open issues I list above).

jsvd:
I'm pretty close to being able to generate the changes of this PR. here's the current diff:

```diff
diff --git a/docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc b/docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc
index 8b72ba7..dd6681a 100644
--- a/docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc
+++ b/docs/versioned-plugins/filters/ruby-v3.1.0.asciidoc
@@ -188,14 +188,5 @@ The path of the ruby script file that implements the `filter` method.
 A key/value hash with parameters that are passed to the register method
 of your ruby script file defined in `path`.
 
-[id="{version}-plugins-{type}s-{plugin}-tag_on_exception"]
-===== `tag_on_exception`
-
-  * Value type is {logstash-ref}/configuration-file-structure.html#string[string]
-  * Default value is `_rubyexception`
-
-Tag to add to events in case the ruby code (either inline or file based)
-causes an exception.
-
 [id="{version}-plugins-{type}s-{plugin}-common-options"]
 include::{include_path}/{type}.asciidoc[]
diff --git a/docs/versioned-plugins/filters/ruby-v3.1.1.asciidoc b/docs/versioned-plugins/filters/ruby-v3.1.1.asciidoc
index 41d9699..6b7e44c 100644
--- a/docs/versioned-plugins/filters/ruby-v3.1.1.asciidoc
+++ b/docs/versioned-plugins/filters/ruby-v3.1.1.asciidoc
@@ -188,14 +188,5 @@ The path of the ruby script file that implements the `filter` method.
 A key/value hash with parameters that are passed to the register method
 of your ruby script file defined in `path`.
 
-[id="{version}-plugins-{type}s-{plugin}-tag_on_exception"]
-===== `tag_on_exception`
-
-  * Value type is {logstash-ref}/configuration-file-structure.html#string[string]
-  * Default value is `_rubyexception`
-
-Tag to add to events in case the ruby code (either inline or file based)
-causes an exception.
-
 [id="{version}-plugins-{type}s-{plugin}-common-options"]
 include::{include_path}/{type}.asciidoc[]
diff --git a/docs/versioned-plugins/filters/ruby-v3.1.2.asciidoc b/docs/versioned-plugins/filters/ruby-v3.1.2.asciidoc
index 375e5a4..12e9377 100644
--- a/docs/versioned-plugins/filters/ruby-v3.1.2.asciidoc
+++ b/docs/versioned-plugins/filters/ruby-v3.1.2.asciidoc
@@ -188,14 +188,5 @@ The path of the ruby script file that implements the `filter` method.
 A key/value hash with parameters that are passed to the register method
 of your ruby script file defined in `path`.
 
-[id="{version}-plugins-{type}s-{plugin}-tag_on_exception"]
-===== `tag_on_exception`
-
-  * Value type is {logstash-ref}/configuration-file-structure.html#string[string]
-  * Default value is `_rubyexception`
-
-Tag to add to events in case the ruby code (either inline or file based)
-causes an exception.
-
 [id="{version}-plugins-{type}s-{plugin}-common-options"]
 include::{include_path}/{type}.asciidoc[]
diff --git a/docs/versioned-plugins/outputs/email-v4.0.4.asciidoc b/docs/versioned-plugins/outputs/email-v4.0.4.asciidoc
index 2ac93cc..c9b5d9c 100644
--- a/docs/versioned-plugins/outputs/email-v4.0.4.asciidoc
+++ b/docs/versioned-plugins/outputs/email-v4.0.4.asciidoc
@@ -20,8 +20,6 @@ include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
-Sends email when an output is received. Alternatively, you may include or
-exclude the email output execution using conditionals.
 
 ==== Usage Example
 [source,ruby]
```

So I'm missing:

- Filling in missing descriptions
- Manual interventions that required adding content like missing documentation for a setting
- The script still performs some hardcoded corrections such as replacing the internal multiline codec link with an external one.

As for the point you mentioned about fixing things only once I completely agree. This sample of 15 allowed me to encode common fixes that we'll see on the 200 plugins, so that if we generate the docs for the 200 plugins you won't have to fix the same thing a million times.
Once docs are generated, fixed and commited I already have a clause that skips a version of a plugin if the asciidoc file exists for it.

dedemorton:
Cool. The first two items in your list (missing descriptions and missing documentation) should be fixed in the latest doc source. We are likely to find other similar problems as we iterate, but it's exciting to see this project coming together!

jsvd:
@dedemorton I pushed this PR into a branch: https://github.com/elastic/logstash-docs/tree/versioned_plugin_docs

We can use this branch as the base branch to build the versioned docs. I created #449, which adds the docs of 14 more codecs on top of this.

PR #461, opened 2018-01-23
final round of plugin doc generation

jsvd:


jsvd:
closed in favor of #462

PR #462, opened 2018-01-23
Final round fixes versioned plugin docs

dedemorton:
Note that for the build to work, this PR also needs to be merged: https://github.com/elastic/logstash/pull/8937


jsvd:
@dedemorton all merged 🎉 thank you for the patience through all the iterations!

PR #463, opened 2018-01-23
auto generated docs for latest plugin versions

jsvd:
@dedemorton this is the result of generating the docs for just the latest version of each plugin.

There are still some things I generate differently, mostly due to lack of information like the description.
My thinking is that we (I) should submit doc PR changes for these 11 plugins, and release new versions.
After this we can start relying on auto generation, or at least script-driven generation.

### DIFF
--- a/docs/versioned-plugins/codecs/compress_spooler-v2.0.6.asciidoc
+++ b/docs/versioned-plugins/codecs/compress_spooler-v2.0.6.asciidoc
@@ -20,7 +20,7 @@ include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
-The compress_spooler codec.
+
 
 [id="{version}-plugins-{type}s-{plugin}-options"]
 ==== Compress_spooler Codec Configuration Options
--- a/docs/versioned-plugins/codecs/csv-v0.1.4.asciidoc
+++ b/docs/versioned-plugins/codecs/csv-v0.1.4.asciidoc
@@ -20,7 +20,7 @@ include::{include_path}/plugin_header.asciidoc[]
 
 ==== Description
 
-The csv codec.
+
 
 [id="{version}-plugins-{type}s-{plugin}-options"]
 ==== Csv Codec Configuration Options
--- a/docs/versioned-plugins/codecs/multiline-v3.0.9.asciidoc
+++ b/docs/versioned-plugins/codecs/multiline-v3.0.9.asciidoc
@@ -24,7 +24,7 @@ The multiline codec will collapse multiline messages and merge them into a
 single event.
 
 IMPORTANT: If you are using a Logstash input plugin that supports multiple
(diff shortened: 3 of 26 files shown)


dedemorton:
@jsvd So is this the full list of changes we need to make to the repos, or is there still some stuff from earlier iterations that needs cleaning up (like the links)? 

Even if the script handles converting the link format from internal to external, we might want to update the links in the repos 'cause people have a way of looking at existing stuff for examples. That wouldn't have to happen before we publish the doc, tho.

jsvd:
> Even if the script handles converting the link format from internal to external, we might want to update the links in the repos 'cause people have a way of looking at existing stuff for examples.

++

@dedemorton I'll dial down the corrections I perform automatically and recompute this PR

jsvd:
@dedemorton skipping doc repair heuristics like the data types link conversion we see that 190 plugins need corrections

dedemorton:
@jsvd Is that all? haha. I'd make the data type conversions an exception, but then...if we do that, we're back to the original problem of people duplicating what they see in the existing dos.

hmmm...so maybe we just leave the doc repair heuristics alone?

jsvd:
@dedemorton I recomputed the PR to apply the data type conversions only. This shows 26 plugins whose latest release tag contains broken asciidoc.

I think it's a reasonable goal to:
a) massively apply the data type link conversions to all plugins
b) fix the remaining broken asciidoc for these 26 plugin
c) mass release plugins with doc changes

WDYT?

jsvd:
https://github.com/logstash-plugins/logstash-codec-compress_spooler/pull/14
https://github.com/logstash-plugins/logstash-codec-csv/pull/2
https://github.com/logstash-plugins/logstash-codec-multiline/pull/61
https://github.com/logstash-plugins/logstash-filter-anonymize/pull/15
https://github.com/logstash-plugins/logstash-filter-fingerprint/pull/36
https://github.com/logstash-plugins/logstash-filter-geoip/pull/135
https://github.com/logstash-plugins/logstash-filter-multiline/pull/34/files
https://github.com/logstash-plugins/logstash-filter-prune/pull/21
https://github.com/logstash-plugins/logstash-filter-split/pull/34
https://github.com/logstash-plugins/logstash-filter-unique/pull/3
https://github.com/logstash-plugins/logstash-filter-uuid/pull/8
https://github.com/logstash-plugins/logstash-input-beats/pull/287
https://github.com/logstash-plugins/logstash-input-file/pull/168
https://github.com/logstash-plugins/logstash-input-lumberjack/pull/84
https://github.com/logstash-plugins/logstash-output-neo4j/pull/5
https://github.com/logstash-plugins/logstash-output-null/pull/6
https://github.com/logstash-plugins/logstash-output-slack/pull/12
https://github.com/logstash-plugins/logstash-output-sns/pull/16

dedemorton:
I'm going to close this PR. I opened issue #9048 to track the remaining tasks for the versioned plugin reference.

PR #478, opened 2018-02-17
Add latest plugin docs

dedemorton:
Updating versioned plugin reference with latest generated files.

The removal of dissect 1.1.3 is OK because 1.1.3 is not a published gem version.


dedemorton:
@karenzone Just an FYI that I've updated the versioned plugin reference and will continue to do so until we have it automated.

PR #491, opened 2018-03-16
Bump Plugin version docs for 6.1.4

robbavey:
Note:

 This commit is larger than I expected - it looks like the plugin documentation was not fully updated to the latest plugin version on previous `6.1` releases.


robbavey:
Thanks @karenzone!

PR #494, opened 2018-03-16
Add latest plugin versions

dedemorton:


dedemorton:
@jsvd Normally I would just merge this directly, but I'm noticing that the logic for generating the overview list doesn't sort the versions correctly when you get version numbers that include double digits, like v5.0.11. Is this something we can tweak?

dedemorton:
Maybe we should sort by date?

jsvd:
@dedemorton I've added [a fix to the versioned docs generator](https://github.com/elastic/docs-tools/commit/7ede134b7732e2f2e0a3ae6276c0133d16e15161) that sorts versions properly. For example, after computing the docs for the beats input:

```
| Version | Release Date
| <<v5.0.11-plugins-inputs-beats,v5.0.11>> | 2018-03-13
| <<v5.0.10-plugins-inputs-beats,v5.0.10>> | 2018-02-21
| <<v5.0.9-plugins-inputs-beats,v5.0.9>> | 2018-02-14
| <<v5.0.8-plugins-inputs-beats,v5.0.8>> | 2018-02-12
| <<v5.0.6-plugins-inputs-beats,v5.0.6>> | 2018-01-05
```

dedemorton:
@jsvd it looks like the fix solved other issues, too, because I see a bunch of versions (with n.n.nn format) that weren't being pulled down before! Can you take a quick look before I merge?

jsvd:
@dedemorton yeah...since it was sorting incorrectly and there were some optimizations to skip content that was already processed, those ended up being ignored.

PR #515, opened 2018-04-17

jsvd:


jsvd:
@dedemorton one more.. it won't be long until https://github.com/elastic/infra/pull/4466 is done

PR #517, opened 2018-04-24
Add bytes filter to the plugin docs

dedemorton:
Do not merge this PR until after the generated plugin docs containing the bytes.asciidoc file have been copied and merged into the relevant repos (master, 6.x, and whichever other 6.n releases require these docs).

@ycombinator The 1.0.0 version of the bytes filter asciidoc breaks the doc build because it references an ID that is not defined anywhere. I think you need to remove this line (or add the missing content) and republish the gem:

`<<plugins-{type}s-{plugin}-digit_group_separator>> |<<string,string>>|No`

You'll need to bump the version, too, so the docgen script can pick up the doc change.

@karenzone This is an example of how you add docs for an entirely new plugin

@jsvd Just a heads up that you'll also have to generate the docs for master (and any other branches where you want this PR) before this PR gets merged

### DIFF
--- a/docs/plugins/filters.asciidoc
+++ b/docs/plugins/filters.asciidoc
@@ -10,6 +10,7 @@ The following filter plugins are available below. For a list of Elastic supporte
 | Plugin | Description | Github repository
 | <<plugins-filters-aggregate,aggregate>> | Aggregates information from several events originating with a single task | https://github.com/logstash-plugins/logstash-filter-aggregate[logstash-filter-aggregate]
 | <<plugins-filters-alter,alter>> | Performs general alterations to fields that the `mutate` filter does not handle | https://github.com/logstash-plugins/logstash-filter-alter[logstash-filter-alter]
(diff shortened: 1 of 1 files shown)


jsvd:
@dedemorton I have created prs to update the docs for master/6.x/6.2 (I think I tested the doc generation correctly against elastic/docs)

ycombinator:
@dedemorton Thanks for catching the non-existent ID issue. I've put up a PR to remove it: 

> You'll need to bump the version, too, so the docgen script can pick up the doc change.

Just so I understand this clearly, I will need to update [this version](https://github.com/logstash-plugins/logstash-filter-bytes/blob/master/logstash-filter-bytes.gemspec#L3), correct? Do I also need to republish the plugin gem at that time?

jsvd:
@ycombinator correct, just bump the patch number and publish. I totally glossed over the need for fixing the docs, which means I'll close the three prs and reopen them after 1.0.1 is out

ycombinator:
@dedemorton @jsvd I just bumped the version for logstash-filter-bytes and published it: https://rubygems.org/gems/logstash-filter-bytes

dedemorton:
I've updated the versioned plugin docs to include the the bytes filter (that way, when this PR is merged, the link to the versioned plugin docs won't break). https://github.com/elastic/logstash-docs/pull/522

PR #521, opened 2018-04-26
Docs bump for 6.3.0

andrewvc:


dedemorton:
Note that I've updated the versioned plugin docs so the links will work

PR #538, opened 2018-05-04
Added link to v4.0.0 google_bigquery documentation

robbavey:
The tagged version of v4.0.0 google big query plugin contains
error which causes the build to fail. Include this correct version
to fix the build

### DIFF
--- a/docs/versioned-plugins/outputs/google_bigquery-v4.0.0.asciidoc
+++ b/docs/versioned-plugins/outputs/google_bigquery-v4.0.0.asciidoc
@@ -1,0 +1,349 @@
+:plugin: google_bigquery
+:type: output
+:default_codec: plain
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: %VERSION%
+:release_date: %RELEASE_DATE%
+:changelog_url: %CHANGELOG_URL%
+:include_path: ../../../../logstash/docs/include
+///////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


docs/versioned-plugins/outputs/google_bigquery-v4.0.0.asciidoc
```diff
@@ -0,0 +1,349 @@
+:plugin: google_bigquery
+:type: output
+:default_codec: plain
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: %VERSION%
```
jsvd: one detail I forgot to mention, we need to fill out these variables. You can see an example here: https://raw.githubusercontent.com/elastic/logstash-docs/versioned_plugin_docs/docs/versioned-plugins/outputs/google_bigquery-v3.2.1.asciidoc

dedemorton:
The script for generating the versioned doc also does some cleanup on the links. Rather than hacking this manually, I'd suggest running the script to generate the docs. Let me know if you want me to do that.

robbavey:
@dedemorton Thanks - can you point me to some instructions? I'd like to be able to do it myself in the event that this needs to be done again...

dedemorton:
@robbavey I had a PR open that contained the step, but removed them because we have a CI job now. I'll send you the steps through email. The other option is to look at the script and make the manual hacks:

https://github.com/elastic/docs-tools/blob/master/versioned_plugins.rb#L200

robbavey:
Thank you @karenzone!

PR #602, opened 2018-07-12
Add azure_event_hubs input to the LS Ref Guide v6.x

karenzone:
Cherrypick from  #576
Note: Generated file is already in place

MERGE TARGET: 6.x


karenzone:
Merged into 6.x and closed

PR #611, opened 2018-08-08
update plugins for 6.4

karenzone:
Manually generated plugin update for 6.4.
Reason: Generate google_pubsub doc so it can be added to Logstash Reference Guide

This is my first time generating the plugin updates, and may be my last now that we have the new automated build process.  Keep me honest!

### DIFF
--- a/docs/plugins/codecs/s3plain.asciidoc
+++ b/docs/plugins/codecs/s3plain.asciidoc
@@ -1,0 +1,33 @@
+:plugin: s3_plain
+:type: codec
+:default_plugin: 0
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: v2.0.7
+:release_date: 2017-12-19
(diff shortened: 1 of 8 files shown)


karenzone:
@dedemorton  I did local builds of everything with no errors.  That doesn't account for old content reappearing.

dedemorton:
@karenzone Right, that makes sense because these's topics probably aren't in the overview files, so they aren't getting built.

jsvd:
It seems the [skip list](https://github.com/elastic/docs-tools/blob/master/settings.yml) hasn't been updated correctly, it is missing:

```
  - logstash-codec-s3plain
  - logstash-filter-anonymize
  - logstash-filter-unique
  - logstash-filter-multiline
```

It could be that folks who have been generated docs edited the skip list locally only (/me is guilty)
@karenzone could you open a PR against docs-tools adding these?

karenzone:
Yes!  I'll open the PR.  Thanks, @dede and @jsvd. I appreciate your help

jsvd:
https://github.com/elastic/docs-tools/pull/23 has been merged so we can regenerate/update these PRs

karenzone:
Closed without merging due to issues with the skiplist. 
Replaced by #614

PR #613, opened 2018-08-08
update plugins for master

karenzone:
Manually generated plugin update for master
Reason: Generate google_pubsub doc so it can be added to Logstash Reference Guide


karenzone:
Closed without merging due to issues with the skiplist and other build issues.
Replaced by #616

PR #619, opened 2018-08-22
Docs for 5.6.11

danhermann:


jsvd:
Interesting, it seems that for 5.6.10 we didn't update our docs, which explains the large amount of changes in plugin docs when 5.6.11 won't actually have none.

That said, it seems we picked up some new plugins (like azure or bytes) that aren't really contained in the release, so we need to remove these.

jsvd:
Overall content LGTM, I didn't try to build the docs against elastic/docs

karenzone:
@jsvd  I had the same question about new plugins, specifically azure and google_pubsub.  I see you struck through that text. Please bring me up to speed on what you're thinking. 

@danhermann and @jsvd  I will build and test.

jsvd:
@karenzone for plugins that aren't included by default it's ok to add the docs, as we also keep documentation about plugins that CAN be installed in that version, so we should keep these

PR #621, opened 2018-09-04
[DISCUSS] First thoughts on doc approach for integrated plugins

karenzone:
I'm putting up this PR to (1) get my head around the implications and possible approaches for integrated plugins docs, and (2) start a discussion with interested parties. 

I've set up a "wrapper" in the LS Ref that points to individual plugin docs.  It points to the results generated from the `index.asciidoc` files present in individual plugin repos.  

I'm putting this out there as a possible Phase 1 (an MVP to get to release quicker). We'd need to do some rework if/when it makes sense to move component docs to an integration repo. Moving all related docs into the integration repo and renaming them would require some rework of our build script.

Notes:
To Do: Requires a PR for //logstash/docs to pull topic into the LS Reference Index. 
To Do: Think through implications of "docs as readme" approach on integrated plugins.

### DIFF
--- a/docs/plugins/integrations.asciidoc
+++ b/docs/plugins/integrations.asciidoc
@@ -1,0 +1,59 @@
+[[integrated-plugins]]
+== Integrated plugins
+
+An integrated plugin combines two or more plugins into one gemspec. 
+This grouping makes it easier to install or update the plugins you need to perform a task.
+Explain more benefits.
+
+These integrated plugins are available:
+
+* <<aws-integrated,AWS>> (https://github.com/logstash-plugins/logstash-integration-aws)
+* <<zeromq-integrated,ZeroMQ>> (https://github.com/logstash-plugins/logstash-integration-zeromq)
+* Another
+* Another
+
+[[aws-integrated]]
+=== AWS integrated plugin
+
+The AWS integrated plugin provides functionality to help you get data into and out of Amazon Web Services (AWS).
+It consists of these individual plugins:
+
+|=======================================================================
+| Plugin | Description 
+| <<plugins-codecs-cloudfront,codec-cloudfront>> | Reads AWS CloudFront reports 
+| plugin-codec-cloudtrail | TBD: Figure out linking. Why is CloudTrail-codec not is LS Ref
+| <<plugins-inputs-cloudwatch,codec-cloudwatch>> | Pulls events from the Amazon Web Services CloudWatch API  
+| <<plugins-inputs-s3,input-s3>> | Streams events from files in a S3 bucket 
+| <<plugins-inputs-sqs,input-sqs>> | Pulls events from an Amazon Web Services Simple Queue Service queue 
+| <<plugins-outputs-cloudwatch,output-cloudwatch>> | Aggregates and sends metric data to AWS CloudWatch 
+| <<plugins-outputs-s3,output-s3>> | Sends Logstash events to the Amazon Simple Storage Service 
+| <<plugins-outputs-sns,output-sns>> | Sends events to Amazon's Simple Notification Service 
+| <<plugins-outputs-sqs,output-sqs>> | Pushes events to an Amazon Web Services Simple Queue Service queue 
+|=======================================================================
+
+
+
+[[zeromq-integrated]]
+=== ZeroMQ integrated plugin
+
+ZeroMQ
+
+input - https://github.com/logstash-plugins/logstash-input-zeromq
+
+filter - https://github.com/logstash-plugins/logstash-filter-zeromq
+
+output - https://github.com/logstash-plugins/logstash-output-zeromq
+
+mixin - https://github.com/logstash-plugins/logstash-mixin-zeromq
+
+
+
+
+
+
+
+
+
+
+
+


karenzone:
@jsvd @robbavey @dedemorton   Here's my attempt to work through some of the doc implications of Integrated plugins, and to start a conversation.  

screen shot 2018-09-04 at 3 41 23 pm

dedemorton:
It feels like we are describing our packaging strategy here, which is fine, but I wonder if we might want to do more than that.

Rather than simply creating more reference-style content that describes our packaging strategy, I wonder if it would make sense to have user-focused, goal-oriented topics that provide some context for when you might want to use a set of plugins together. Then you would have the opportunity to add value beyond just saying that the plugins can be installed together. I'm guessing there are times when an integrated plugin will include filters or codecs...and it might not be obvious why they're included.

jsvd:
++ on figuring out what should the overall table of contents structure should be from the perspective of what the user would like to see.

Thinking about what the user wants, would be something like:
"I have data in kafka that I want to use" or "I want to send data to rabbitmq"
Then you'd go to logstash documentation, and see:

```
Logstash Reference Docs:
- Introduction
- Getting Started with Logstash
- How Logstash Works
- Setting Up and Running Logstash
- Logstash Integrations
   \+ Amazon AWS
   |- Elasticsearch
       + Input
       + Filter
       + Output
   |- Kafka
       + Input
       + Output
   |+ RabbitMQ
- Transforming Data
- Managing Logstash
```

Clicking on "Logstash Integrations would show a documentation page similar to this PR's docs/plugins/integrations.asciidoc

Clicking on "RabbitMQ" would jump to a page about the integration, and list the plugins that compose the integration, with links that would jump to the same place as the "RabbitMQ -> Input", for example.

The documentation for each plugin under the integration would be pretty much the current documentation for each plugin.

Now, thinking about document structure on github, the integration X documentation landing page would live under "logstash-integration-X/docs/index.asciidoc". The documentation for each plugin within the integration could be in either:
- logstash-integration-X/docs/input-rabbitmq-index.asciidoc
- logstash-integration-X/docs/input/rabbit-index.asciidoc

It can't just be "input" as an integration can have multiple input, filter or output plugins (e.g. S3 and SQS inputs for the AWS integration).

WDTY?

tbragin:
@acchen97^^ FYI

acchen97:
I'm trying to understanding what's the value in documenting these integrated plugins together as separate sections. This solves a packaging problem that helps us optimize as developers, but from a user standpoint, these are quite different use cases even within an integration plugin. For example, most users will only care about the ES output for the integrated ES plugin. The ES filter is for a completely different use case w.r.t lookup enrichment. Additionally, a user who's interested in Cloudwatch/trail inputs may not be interested in SNS/SQS outputs. Oftentimes when a user comes to our docs, they'll likely want to answer questions that seem to be nicely addressed with the way our plugin docs are structured today: what data sources can I ingest from, what processing or enrichment capabilities are available, what data destinations can I persist to.

If we weren't planning to bundle these integrated plugins by default, I could see value in documenting them together so users know what they're installing as part of this "pack". If this new packaging doesn't change anything from a plugin usage perspective, perhaps it could make sense to leave the docs structured the way they are. @jsvd how does this affect how plugins are installed/updated? Can we do a single plugin update, or is it all (i.e. always updates all plugins within the ES integrated plugin) or nothing?

jsvd:
> I'm trying to understanding what's the value in documenting these integrated plugins together as separate sections.

The goal here was to make the information about integrating logstash with an external service easier to find. If the user wants to browse the list of plugins, the inner plugins would still show up in the list as we have it now. But if the user wants to learn about what their options are when interacting with ES, it's easier to show that the ES integration contains a input/filter/outputs, than hoping the user extrapolates "ok logstash has an ES output, maybe there's an input or filter too?"

> how does this affect how plugins are installed/updated? Can we do a single plugin update, or is it all (i.e. always updates all plugins within the ES integrated plugin) or nothing?

You upgrade the whole integration as a single plugin (e.g. `bin/logstash-plugin update logstash-integration-elasticsearch`). As the plugins within the integration will grow in coherence and reuse more code, this will be necessary.

acchen97:
@jsvd got it. These are the times when I wish we had a dedicated "integrations" page. :) A potential concern I have with an "integrated plugins" section like this is that our current spread of integrations is small, so a user may look at this and think we have AWS integrations, but not GCP integrations. They could look into our input/output plugins to find them, but it could still lead to misunderstanding. Thoughts?

jsvd:
@karenzone I've spoken with @acchen97 and I agree with the idea to minimize the impact of the integration plugins in the docs. The idea would then be to, first and foremost, allow the inner plugins of an integration to show up in the existing input/codec/filter/output sections.
The change here would be that the headers for each inner plugin that show the version and release date would now show the integration version/date instead. There would also be a link to a new page that describes the integration and other inner plugins contained in it.
Let's take the example of the rabbitmq plugins. The documentation for the [rabbitmq input](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-rabbitmq.html) currently shows:

```
Rabbitmq input plugin
Plugin version: v6.0.3
Released on: 2018-04-06
Changelog
```

This would change to, when we migrate to the [integration plugin](https://github.com/jsvd/logstash-integration-rabbitmq):
```
Rabbitmq input plugin
Plugin version: v1.0.0         # version of the rabbitmq integration
Released on: 2018-10-02  # release date of the rabbitmq integration version
Changelog    # changelog for the integration

This plugin is part of the [RabbitMQ integration](link to integration).
```

The "link to integration" would be a new documentation page that shows

```
Logstash RabbitMQ integration

This integration is for connecting to rabbitmq. There's a plugin to read from RabbitMQ and one to send events to RabbitMQ.

- Logstash Input Rabbitmq (w/ link back to the documentation)
- Logstash Output Rabbitmq (w/ link to output plugin documentation page)
```

These pages that describe the integrations would all be listed in a new reference doc section close to the bottom of [the table of contents](https://www.elastic.co/guide/en/logstash/6.4/index.html), below "Codec Plugins" and above "Contributing to Logstash".

@karenzone let me know what you think about this. To sum it up, the idea is to keep existing flow of searching for docs by plugin type with a little extra information that some of the plugins are grouped in integrations.

PR #634, opened 2018-10-26
Add cloudtrail codec docs back

robbavey:
Cloudtrail codec documentation was previously removed because it was unable to be installed in later versions of Logstash. This has now been resolved, so the documentation can be re-added.


karenzone:
UPDATE:  https://github.com/elastic/logstash/pull/10190 removed codec-cloudtrail from the skiplist. Next time we generate plugin docs for master, 6.x, 6.5, 6.4, codec-cloudtrail should be included.
TO DO:
- [ ] Kick off plugin docs builds for various repos
- [ ] Verify that `codecs/cloudtrail.asciidoc` if present for all repos
- [ ] Test and merge this PR

PR #636, opened 2018-10-27
Update docs master and add no_codec tags

karenzone:
Generated plugins from #632 with no_codec tags added for input-snmp and output-appsearch


karenzone:
Closed without merge

PR #638, opened 2018-11-01
Add conditional coding to support plugins with no_codec

karenzone:
Conditional coding was added to LS Ref in https://github.com/elastic/logstash/pull/10078.  Duplicating change here to support no_codec option in Versioned Plugin Reference.


karenzone:
DO NOT MERGE:  This change is blowing up local builds. 

@dede  SHOULD this build from my branch from `elastic:versioned_plugin_docs`? I'm not sure because of all of the VPR magic behind the scenes.

PR #650, opened 2018-11-13
Update docs 6.x nocodec

karenzone:
Generated plugins for 6.x (#649) with no_codec tags added for output-elastic_app_search


karenzone:
Update:  Unassigned JARVIS.  I found something else I want to fix.

PR #651, opened 2018-11-13
updated docs for master

logstashmachine:


karenzone:
Note to self: The codec is incorrect, but will be updated the next time we bump the plugin version

PR #661, opened 2019-01-16
Fix syntax in dictionary example for filter-translate - 5.6

karenzone:
Remove commas from `dictionary` example


yaauie:
As mentioned elsewhere, since we're pushing this docs fix to 5.6, we should also target the intermediate branches.

PR #667, opened 2019-01-29
Add instructions to upgrade the plugin to use ILM

robbavey:

PR #672, opened 2019-01-30
Updated Plugin docs for 6.6 plus new TOC entries for filter-http and filter-memcached

karenzone:
Starts with generated docs for 6.6  #671 and adds:
- new logstash-filter-http to Logstash Reference Guide
- new logstash-filter-memcached to Logstash Reference Guide
- ILM note from #667 to logstash-output-elasticsearch


karenzone:
@robbavey FYI, I manually added the text we added today with #667

PR #681, opened 2019-03-26
Updated plugin docs for 6.7

robbavey:
Built manually due to Jenkins issues

### DIFF
--- a/docs/plugins/filters/geoip.asciidoc
+++ b/docs/plugins/filters/geoip.asciidoc
@@ -5,9 +5,9 @@
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
 ///////////////////////////////////////////
-:version: v6.0.0
-:release_date: 2019-01-29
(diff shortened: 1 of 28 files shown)


docs/plugins/filters/geoip.asciidoc
```diff
@@ -44,7 +44,7 @@ A `[geoip][location]` field is created if
 the GeoIP lookup returns a latitude and longitude. The field is stored in
 http://geojson.org/geojson-spec.html[GeoJSON] format. Additionally,
 the default Elasticsearch template provided with the
-{logstash-ref}/plugins-outputs-elasticsearch.html[elasticsearch output] maps
+<<plugins-outputs-elasticsearch,`elasticsearch` output>> maps
```
jsvd: I think we need to use the logstash-ref style urls here?

karenzone: @jsvd  Good catch! You are correct. I'll fix it here and then fix it in the source.

PR #682, opened 2019-03-29
[Doc] Prototype landing page for partner plugin

karenzone:
Prototype to spark conversations. That is, you'll see some placeholders and inconsistencies. 
https://preview-docs.firebaseapp.com/partner-plugin-proto/filter-plugins.html


karenzone:
@acchen97 @jsvd  Prototype we discussed. Rendered here: https://preview-docs.firebaseapp.com/logstash-ref/filter-plugins.html

acchen97:
@karenzone thanks, I'm still waiting on a description from the partner. Stay tuned...

acchen97:
@karenzone I like the concept here a lot, thanks for the prototype. Although this is slightly different from our other plugins, I think it would be good to have a "Getting Help" section like our other plugins. Also, would it make sense to include the latest plugin version given we can keep this dynamic as updates come?

karenzone:
@acchen97  Clarification, please.
> would it make sense to include the latest plugin version given we can keep this dynamic as updates come?

True that the changelog and repo links will always point to the latest versions.  Besides that, how do you see us knowing when the plugin is updated on their side?  In other words, if we added the plugin version, how would we keep that info fresh? 

Thanks for the comments and suggestions.

karenzone:
@acchen97  One more clarification needed:

> I think it would be good to have a "Getting Help" section like our other plugins.

Here's the "Getting Help" content for plugins that live in the logstash-plugins org.  
> "For questions about the plugin, open a topic in the Discuss forums. For bugs or feature requests, open an issue in Github. For the list of Elastic supported plugins, please consult the Elastic Support Matrix."

Should we modify it for outside plugins? I can set the github link to target the empow-classifier repo if they are open to that.  Is suggesting "Discuss" the right approach for an outside plugin?  Does it make sense to include the link for Elastic supported plugins? 

Thanks!

acchen97:
> True that the changelog and repo links will always point to the latest versions. Besides that, how do you see us knowing when the plugin is updated on their side? In other words, if we added the plugin version, how would we keep that info fresh?

Perhaps pulling Rubygems metadata? If its too hairy, its ok if we leave it out.

> Should we modify it for outside plugins? I can set the github link to target the empow-classifier repo if they are open to that. Is suggesting "Discuss" the right approach for an outside plugin? Does it make sense to include the link for Elastic supported plugins?

Agreed we should simplify the verbiage here. How about this: "This is a third-party plugin. For bugs or feature requests, open an issue in Github."

Looking again, we can probably add the "Installation" section similar to our other plugins as well. This plugin will always need to be installed prior to usage, so maybe just: "It is easy to install this plugin by running bin/logstash-plugin install logstash-filter-.... See Working with plugins for more details." Would you agree?

karenzone:
@acchen97 Thanks for the good comments! I'll get them in and re-post on firebase

acchen97:
@karenzone thanks for updating the preview.

I think we're close. The suggestions I made in the [last comment](https://github.com/elastic/logstash-docs/pull/682#issuecomment-489930244) are still outstanding. Beyond that, I think it makes sense to remove the "Gem" link for consistency with our other plugins.

karenzone:
Firebase link (updated): https://preview-docs.firebaseapp.com/partner-plugin-proto/filter-plugins.html

acchen97:
LGTM thanks @karenzone

PR #679, opened 2019-03-14
Add google_cloud_storage to list of inputs

karenzone:
Adds new input-google_cloud_storage to LS Ref.

IMPORTANT: Generated `inputs/google_cloud_storage.asciidoc`  files  must be present before this PR can be merged into a branch.  Otherwise, the doc build will fail.

Branch 7.0
- [x] file present 
- [x]  PR merged

Branch 7.1
- [x] file present 
- [x]  PR merged

Branch 7.2
- [x] file present 
- [x]  PR merged

Branch 7.3
- [x] file present
- [ ]  PR merged

Branch 7.x
- [x] file present
- [ ]  PR merged

Branch master
- [x]  file present
- [ ]   PR merged

PR #683, opened 2019-04-04
Manual build of 7.0 versioned plugin docs

robbavey:


robbavey:
@jsvd, @karenzone Manual build of 7.0 versioned plugin docs - can you make sure that all the expected plugins are there (and no extra ones)? I had to use an old version of the `Gemfile`, as the latest `Gemfile.template` doesn't include default plugins, which causes the gradle task to generate the json marks most of the plugins as non default.

karenzone:
I verified that everything builds correctly and looks good.  I'm glad to see google_cloud_storage is present.  After this PR is merged, I'll be able to merge #679 *for 7.0 branch*.

jsvd:
@robbavey we added these plugins to the package, can you get their docs in as well?
> logstash-codec-avro
> logstash-filter-prune
> logstash-filter-uuid
> logstash-input-couchdb_changes

robbavey:
@karenzone @jsvd Changes are up

karenzone:
Still builds cleanly and still LGTM

PR #686, opened 2019-04-16
Clean up formatting in older plugin docs

karenzone:
Clean up formatting to enable converting Versioned Plugin Reference from asciidoc to asciidoctor. 
NOTE:  The formatting is correct in the plugin repos and in current versions.


karenzone:
@nik9000 Here's the asciidoc cleanup work you suggested for the VPR. I've fixed content in the source files in the plugin repos as well, so that new content coming over should be clean.

karenzone:
@nik9000  That would be awesome!
I tagged @jsvd to be sure he's in the know, but it builds cleanly in docker. Likely nothing here that would prevent us from merging.

PR #688, opened 2019-04-17
Asciidoctor support

nik9000:
This fixes a few Asciidoctor specific issues with the docs, specifically
removing an `and` from an `ifeval` which is not supported by AsciiDoc
and adding some substitutions that are not the default for Asciidoctor.

### DIFF
--- a/docs/versioned-plugins/include/6.x/plugin_header.asciidoc
+++ b/docs/versioned-plugins/include/6.x/plugin_header.asciidoc
@@ -1,9 +1,11 @@
 ifeval::["{versioned_docs}"!="true"]
+[subs="attributes"]
 ++++
 <titleabbrev>{plugin}</titleabbrev>
 ++++
 endif::[]
 ifeval::["{versioned_docs}"=="true"]
+[subs="attributes"]
 ++++
 <titleabbrev>{version}</titleabbrev>
 ++++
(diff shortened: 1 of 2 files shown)


karenzone:
@nik9000  I can get this to build successfully with both docker/asciidoc and docker/asciidoctor.  I noticed that with --asciidoctor, all of the versions of a single plugin appear on one page.  Is that something we should look at now or in a later iteration? 
@jsvd  FYI
@dedemorton  The code changes look straightforward, but I'd appreciate your taking a look since this magic of the VPR is your creation.

nik9000:
> I noticed that with --asciidoctor, all of the versions of a single plugin appear on one page. Is that something we should address now or resolve in a later iteration?

Hmmmm. I think we should grab it in a follow up change. This is progress and progress is good.

docs/versioned-plugins/include/6.x/version-list-intro.asciidoc
```diff
@@ -1,6 +1,7 @@
 [id="{type}-{plugin}-index"]
```
dedemorton: ditto. this is copied from the logstash repo. Make sure you also apply the changes to the file in the logstash repo, too.

docs/versioned-plugins/include/6.x/plugin_header.asciidoc
```diff
@@ -1,9 +1,11 @@
 ifeval::["{versioned_docs}"!="true"]
+[subs="attributes"]
```
dedemorton: Note that this file is a manual copy of https://github.com/elastic/logstash/blob/master/docs/include/plugin_header.asciidoc. Make sure you also apply the changes to the file in the logstash repo. Someday the plugin script will copy the file automatically (gotta dream big), but for now, we have to keep the file in sync manually.

nik9000:
Awesome! I'll merge this today and see about opening the changes to the logstash repo too!

nik9000:
I've opened https://github.com/elastic/logstash/pull/10730 .

PR #723, opened 2019-05-27
updated docs for 7.2

logstashmachine:


docs/plugins/filters/oui.asciidoc
```diff
@@ -0,0 +1,71 @@
+:plugin: oui
+:type: filter
+:default_plugin: 0
```
karenzone: New plugin file. Do we want to link it from list of filters?

docs/plugins/inputs/github.asciidoc
```diff
@@ -1,14 +1,13 @@
...
 
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
 ///////////////////////////////////////////
-:version: v3.0.7
-:release_date: 2018-04-06
-:changelog_url: https://github.com/logstash-plugins/logstash-input-github/blob/v3.0.7/CHANGELOG.md
+:version: v3.0.5
+:release_date: 2017-11-07
+:changelog_url: https://github.com/logstash-plugins/logstash-input-github/blob/v3.0.5/CHANGELOG.md
```
karenzone: Verify that this regression is correct. It removes `:default_codec!:`, and I don't think that's what we want.

docs/plugins/filters/oui.asciidoc
jsvd: ++ on adding it

docs/plugins/inputs/github.asciidoc
jsvd: it is, because a transitive dependency causes a previous version to be installed. we may need to sort this out but it is not blocking for this merge. 
Regarding the codec, this plugin doesn't use codecs. we can introduce the no_codec setting in a separate pr

jsvd:
this is now outdated, closing.

PR #727, opened 2019-06-17
updated docs for 7.2

logstashmachine:

PR #731, opened 2019-06-21
Add oui to list of filters

karenzone:
Adds new oui filter to list of filter plugins in LS Ref.

IMPORTANT: Generated `filters/oui.asciidoc` file must be present before this PR can be merged into a branch. Otherwise, the doc build will fail. 

Branch `7.2`
- [x] filters/oui.asciidoc file is present (#727)
- [ ] PR merged

Branch `7.x`
- [ ] filters/oui.asciidoc file is present
- [ ] PR merged

Branch `master`
- [ ] filters/oui.asciidoc file is present
- [ ] PR merged


yaauie:
What filter is this that we are adding? I'm finding [logstash-plugins/logstash-filter-oui](https://github.com/logstash-plugins/logstash-filter-oui) which has outstanding issues indicating that it may not be compatible with Logstash 6.x (and therefore also 7.x, I would assume).

If that is the one, I would prefer we get it up to snuff before linking to it in docs (although I'm unsure what resources we have to do so at the moment).

karenzone:
Thanks for the feedback, @yaauie.  The unlinked asciidoc file was generated and picked up in #723 (even though it was merged with #727 instead).  I asked about adding it and got ++.

karenzone:
Closing. Plugin needs some work before we add it to the docs.

PR #734, opened 2019-06-27

logstashmachine:


karenzone:
https://github.com/elastic/logstash/pull/10921 should fix the build problem

PR #735, opened 2019-06-28
Updated docs for 7.2

robbavey:
This was the result of the manual docs build for 7.2. Note that some unexpected changes were included - none of the updates to filter plugins included here are currently in 7.2

This PR is a draft PR for discussion

### DIFF
--- a/docs/plugins/filters/http.asciidoc
+++ b/docs/plugins/filters/http.asciidoc
@@ -5,9 +5,9 @@
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
 ///////////////////////////////////////////
-:version: v1.0.2
-:release_date: 2019-06-24
-:changelog_url: https://github.com/logstash-plugins/logstash-filter-http/blob/v1.0.2/CHANGELOG.md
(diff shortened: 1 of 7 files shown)


karenzone:
@robbavey  Thank you for generating the output and putting this out for discussion.

karenzone:
@robbavey @jsvd  Do we know if the lock files are working as expected in this scenario?  If so, my understanding of their role is not accurate, and I need an update. 

What's our "go forward" plan?  If we need more time to look into this issue, I can branch this PR and put up just the files I need right now -- `output-google_cloud_storage` and `input-jms`.  I also need the fix that's in `filter-kv` and I'm not sure why that's not breaking stuff.  The version number for filter-kv is higher than the lock file, so I might need to make the fix manually.

jsvd:
> Do we know if the lock files are working as expected in this scenario? If so, my understanding of their role is not accurate, and I need an update.

It seems that since we removed the list of default plugins from the Gemfile.template the versions of the default plugins in the lock file template are being ignored. Further investigation is necessary..until then we can skip the changes for default plugins, which will require some manual work.

PR #751, opened 2019-07-24
Fix experimental flag for asciidoctor

nik9000:
Asciidoctor isn't able to deduce the right version of the experimental
tag when it is written on its own line. This moves it to the line below.


nik9000:
This is how it looks to me after I switch to this branch:

2019-08-22-154744_1576x225_scrot

karenzone:
@elasticmachine, run elasticsearch-ci/docs

PR #762, opened 2019-07-31
updated docs for 7.3

logstashmachine:


jsvd:
@karenzone we're back on :)

PR #782, opened 2019-08-08
[Backport] Temp fix for broken link

karenzone:
Backport of #780.

Beats troubleshooting docs were restructured, causing a doc build break in `master`. The doc build will break again when the Beats auto-backports happen.  This PR removes the deep links to avoid doc build breaks in 7.x and 7.3. 

After Beats autobackports happen and the dust settles, the correct fix is here: #781.


karenzone:
Merged for 7.x and 7.3.

PR #786, opened 2019-09-23
update docs for 7.4.0

jsvd:


docs/plugins/inputs/file.asciidoc
```diff
@@ -380,7 +380,7 @@ The sincedb record now has a last active timestamp associated with it.
 If no changes are detected in a tracked file in the last N days its sincedb
 tracking record expires and will not be persisted.
 This option helps protect against the inode recycling problem.
-Filebeat has a {filebeat-ref}/faq.html[FAQ about inode recycling].
+Filebeat has a {filebeat-ref}/faq.html#inode-reuse-issue[FAQ about inode recycling].
```
jsvd: @karenzone this seems to be causing the docs build to fail, need your help here :)

karenzone: run elasticsearch-ci/docs

karenzone: @elasticmachine, run elasticsearch-ci/docs

karenzone: Note to self (@karenzone):  Everything looks good with the exception of the four plugins I didn't publish before this update was created. I think we should merge this PR, and then immediately follow up with another docgen to pick up the latest.

PR #787, opened 2019-09-25
update docs in master

jsvd:


jsvd:
@karenzone now that we have the pr docs check, can we start thinking of merging these automatically if they pass CI?

karenzone:
> @karenzone now that we have the pr docs check, can we start thinking of merging these automatically if they pass CI?

Let's discuss. We could merge without breaking the builds, and that would certainly keep things moving along.  However, we often introduce regressions that won't break the build, but do affect the quality of the docs.

PR #788, opened 2019-10-10
Generate plugin docs for master

karenzone:
Updated docs to pick up recently published plugins


karenzone:
The previous docs update #787 wasn't complete because I had not published four plugins.

PR #791, opened 2019-10-16
updated docs for master

logstashmachine:


karenzone:
This plugin is not currently in the Logstash Reference Guide[https://www.elastic.co/guide/en/logstash/master/output-plugins.html].  Should we add it?

karenzone:
Decision: The docs are incomplete, and so we will not add.  
Added to skiplist:  https://github.com/elastic/docs-tools/pull/35

PR #793, opened 2019-10-18
Update plugin docs for 7.3

karenzone:
Updated plugin docs (manually generated)

Notes to self:  
* The manual docgen process was wonky and threw lots of errors. I finally got some output, but I don't want to assume that the output is correct without a detailed review.
* The script picked up `output-monasca_log_api` as changed. Should I add it?

Lockfile: https://github.com/elastic/logstash/blob/7.3/Gemfile.jruby-2.5.lock.release


karenzone:
UPDATE:  Output looks squirrelly.  DO NOT MERGE. 
Some of the versioning does not match what's expected based on lockfile. More troubleshooting is in order.

PR #783, opened 2019-08-12
updated docs for 7.3

logstashmachine:

PR #794, opened 2019-10-22
Update plugins 7.3 plus

karenzone:
This PR starts with the goodness in #783, and then:
* fixes a link that would break the doc build (`inputs-file`)
* removes some files that would introduce regressions to asciidoctor formatting (`outputs-circonus`, `outputs-librato`, `outputs-riak`, `outputs-riemann`, and `outputs-google_bigquery`.


yaauie:
LGTM 👍

PR #812, opened 2019-12-02
[DOCS] Tweaks to integration plugins index page

jsvd:


docs/plugins/integrations.asciidoc
```diff
@@ -1,17 +1,13 @@
...
 | <<plugins-integrations-kafka,kafka>> | Plugins for use with the Kafka distributed streaming platform. | https://github.com/logstash-plugins/logstash-integration-kafka[logstash-integration-kafka]
 | <<plugins-integrations-rabbitmq,rabbitmq>> | Plugins for processing events to or from a RabbitMQ broker. | https://github.com/logstash-plugins/logstash-integration-rabbitmq[logstash-integration-rabbitmq]
 |=======================================================================
 
-:edit_url: https://github.com/logstash-plugins/logstash-integration-kafka/edit/master/docs/index.asciidoc
-include::integrations/kafka.asciidoc[]
-
-:edit_url: https://github.com/logstash-plugins/logstash-integration-rabbitmq/edit/master/docs/index.asciidoc
-include::integrations/rabbitmq.asciidoc[]
+:edit_url: https://github.com/elastic/logstash-docs/edit/master/docs/plugins/integrations.asciidoc[]
```
karenzone: Should stay as is.  This pulls in the asciidoc files for kafka and rabbitmq, and correctly sets the edit url overrides.

docs/plugins/integrations.asciidoc
```diff
@@ -1,17 +1,13 @@
 [[plugin-integrations]]
-== Plugin integrations
+== Integration plugins
```
karenzone: What about "*Integrated* plugins."  "Integration plugins" sounds like a list of plugins that you'd use for integrations.

jsvd: we already introduced the concept in a previous article: https://www.elastic.co/blog/logstash-lines-introduce-integration-plugins.

> sounds like a list of plugins that you'd use for integrations.

It's indeed something like that, you use the integration plugins to integrate with  kafka or rabbitmq, (or jdbc/elasticsearch/aws).

docs/plugins/integrations.asciidoc
jsvd: My reason for this edit is that clicking the "edit" button in the [integrations list page](https://www.elastic.co/guide/en/logstash/7.5/plugin-integrations.html) jumps into rabbitmq  docs, which is confusing.

karenzone: You have uncovered a weird behavior change from our conversion to asciidoctor. @nik9000 has discovered that html5 fixes the edit_me URL issue and has [added a test](https://github.com/elastic/docs/pull/1509).

docs/plugins/integrations.asciidoc
jsvd: Hah..words are hard..We have used the term integrations to have umbrella plugins dedicated to integrating with a technology (which may have inputs/filters/outputs, etc). It's true that an elasticsearch output integrates with elasticsearch but the elasticsearch integration plugin will provide all that you need for this integration.

In a perfect world you'd have integration plugins that bundle inputs/filters/outputs, and all the transformation plugins like mutate, clone, kv, grok, geoip would be the non-integration plugins.

karenzone: I understand the concept, and I understand that precedent wins.  :-)
I'm OK with the way you have.

PR #818, opened 2020-01-06
Manual plugin generation with formatting fixes for 6.8

karenzone:
Manual generation for plugin docs with formatting fixes to filter-memcached


karenzone:
RESEARCH:  The manual plugin generation process picked up google_cloud_storage as new for 6.8. Should it be added to 6.8 index?

karenzone:
Note to self (@karenzone): I intentionally regressed link to beats faq because the faq reorg happened for 7.x branches.

PR #813, opened 2019-12-10
Add reference to JDBC integration plugin

robbavey:
Adds link and reference to the JDBC integration plugin to the
integration plugin index page.

PR #827, opened 2020-01-13
Add reference to JDBC integration plugin (Backport of #813)

robbavey:
* Add reference to JDBC integration plugin

Adds link and reference to the JDBC integration plugin to the
integration plugin index page.

* Fix typo

Backport of #813


robbavey:
Thank you @karenzone!

PR #834, opened 2020-02-06
Plugin docs manually generated for 6.8

karenzone:
Plugin docs for 6.8 (manually generated)

### DIFF
--- a/docs/plugins/filters/memcached.asciidoc
+++ b/docs/plugins/filters/memcached.asciidoc
@@ -111,7 +111,7 @@ filter {
   * Value type is <<hash,hash>>
   * There is no default value for this setting.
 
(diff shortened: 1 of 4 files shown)


kares:
hopefully, doc-gen is better on 7.x branches - if not we might want to take a look.

karenzone:
@kares thanks for the review!

> hopefully, doc-gen is better on 7.x branches - if not we might want to take a look.

The new PR docs checker is awesome, but it causes failures on bad links and/or bad asciidoc formatting.  Those problems exist on 6.8 because memcached-filter is locked at an old version and because the link that file-input targets wasn't updated for 6.8.  (I've fixed memcached-filter formatting in the docs output 10+ times.) 

Similar problems could theoretically manifest in the 7.x branches, but aren't  as likely if we do good doc reviews.

PR #852, opened 2020-04-13
[Doc]Rename java_sink plugin

karenzone:
This PR and https://github.com/elastic/logstash/pull/11778 are dependent on each other. We need to synchronize merging for all branches.


roaksoax:
lgtm!

PR #864, opened 2020-05-06
updated docs for 6.8

logstashmachine:


jsvd:
@karenzone I've respun the 6.8 doc job that failed (I think it is an issue with the centos 8 worker that jenkins started up). I think we can close this PR as it would regress the fix we did on the file input docs.

karenzone:
Closed without merging for reasons stated above.  Net no new changes.

PR #884, opened 2020-07-15
Doc:Fix malformed link to get docs building successfully

karenzone:


karenzone:
Yes, there is, but the doc build for the entire company is broken.  This PR is a quick fix to get the doc builds going again while I fix it properly.

PR #892, opened 2020-07-27
Use shield attribute from shared legacy attributes

gtback:
This URL is going to change soon. See elastic/docs#1918

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,13 +1,14 @@
 [[logstash-reference]]
 = Reference
 
+include::{docs-root}/shared/legacy-attrs.asciidoc[]
+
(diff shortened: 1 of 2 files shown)


karenzone:
@elasticmachine, run elasticsearch-ci/docs

gtback:
@elasticmachine run elasticsearch-ci/docs

gtback:
@karenzone Oh, this one is showing `{shield-13}`. Is it possible that it won't work until both this and elastic/logstash#12147 are merged? I'm willing to merge this one in and then retry the other one if you're OK with that (the builds are at least passing, and it's in really old versions so I doubt anyone is relying on those links).

PR #893, opened 2020-08-05
Doc:Fix links to es broken at 7.9 cutover

karenzone:
Elasticsearch doc improvements and restructuring broke some Logstash links at 7.9 and before.  

This problem is recurring because we use `/current/`  in links as a work around for versioning incompatibilities between the stack and individual plugins.   When `current` resolves to a new branch, any link targets that no longer exist are broken.  For example, if a link exists in 7.7 but not in 7.8, it will break when `/current/`=7.8.  This situation is a nightmare when it happens on release day and the docs build breaks. 

In this PR I added links to more appropriate versions. These files are generated, and so fixing the output should make for a more evergreen fix over just changing the link to resolve for `current`.  

There's more to be done, and I'll make those changes in plugin source. 

**PREVIEWS:** https://logstash-docs_893.docs-preview.app.elstc.co/diff

### DIFF
--- a/docs/versioned-plugins/inputs/elasticsearch-v4.3.0.asciidoc
+++ b/docs/versioned-plugins/inputs/elasticsearch-v4.3.0.asciidoc
@@ -166,7 +166,7 @@ It will be removed in the next major version of Logstash.
 If document metadata storage is requested by enabling the `docinfo`
 option, this option lists the metadata fields to save in the current
 event. See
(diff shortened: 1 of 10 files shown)


karenzone:
@jsvd. I'd like to sync with you to explain what I did and why, and to talk about next steps.

jsvd (approved):
LGTM

PR #900, opened 2020-08-20
auto generated update of versioned plugin documentation

logstashmachine:

PR #905, opened 2020-08-26

logstashmachine:


karenzone:
@jsvd:  Looks like I was supposed to merge #900 (instead of closing it) even though it returned FAILURE. This autogen job didn't pick up those files.

jsvd:
let's see what happens if I close and remove the branch.

PR #907, opened 2020-08-31
Doc:Update split filter description in Filters index

karenzone:
Related: https://github.com/logstash-plugins/logstash-filter-split/pull/43


karenzone:
@colinsurprenant This is a follow-up fix to the one you helped me with in https://github.com/logstash-plugins/logstash-filter-split/pull/43. Should be an easy review.  Will you give it a look, please?

PR #943, opened 2020-09-28
Doc:Move sink output to correct order

karenzone:
`java-sink` was renamed to `sink`. This work corrects ordering in table list
and table of contents.

**PREVIEW:** https://logstash-docs_943.docs-preview.app.elstc.co/guide/en/logstash/master/output-plugins.html


andsel (approved):
LGTM, thanks @karenzone

PR #949, opened 2020-10-01
updated docs for 6.8

logstashmachine:


karenzone (commented):
3 of 4 files in this PR contain regressions we don't want to take.
Changes to filter-cipher are valid.

karenzone:
Closed without merging. Replaced by #950

PR #950, opened 2020-10-01
Updated docs for 6.8 with regressions removed

karenzone:
This work starts with generated plugin docs (#949), and removed files containing regressions


karenzone (commented):
Builds cleanly and renders correctly with regressions removed.

PR #948, opened 2020-10-01
updated docs for master

logstashmachine:

PR #951, opened 2020-10-02
Updated docs for master with regressions removed

karenzone:
This work starts with generated plugin docs (#948), and removes files containing regressions


karenzone:
A mix of commits and branches from 7.9 and master. Closing

PR #946, opened 2020-10-01
updated docs for 7.9

logstashmachine:

PR #952, opened 2020-10-02
Updated docs for 7.9 with regressions removed

karenzone:
This work starts with generated plugin docs (#946 ), and removes files containing regressions


karenzone (commented):
Builds cleanly and renders correctly with regressions removed.

PR #953, opened 2020-10-07
updated docs for 7.10

logstashmachine:


karenzone (changes requested):
Regression to `docs/plugins/inputs/log4j.asciidoc`.  
I'll remove the updates and commit.

karenzone (approved):
Builds cleanly and renders correctly with log4j regressions removed.

PR #954, opened 2020-10-28
updated docs for master

logstashmachine:


karenzone (commented):
Changes picked up for `docs/plugins/inputs/log4j.asciidoc` are a regression, and are causing doc build failures.

karenzone (approved):
Builds cleanly and renders correctly with log4j regressions removed.

PR #958, opened 2020-11-03
Set branch for es-input in Versioned Plugin Reference

karenzone:
Our use of `current` in the plugin docs frequently causes problems when we cut over to a new minor or major version.  

This PR: 
* fixes links that would be broken for 7.10
* sets an appropriate stack `branch` to "future proof" against broken links

**PREVIEW:** https://logstash-docs_958.docs-preview.app.elstc.co/diff

Notes:
* 4.71 -> 7.9
* 4.8.0 -> 7.10
* 4.8.1 -> 7.10


yaauie (approved):
LGTM 👍

PR #960, opened 2020-11-03
Set branch for es-output in Versioned Plugin Reference

karenzone:
Our use of `current` in the plugin docs frequently causes problems when we cut over to a new minor or major version.

This PR:

* fixes links that would be broken for 7.10
* sets an appropriate stack branch to "future proof" against broken links

PREVIEW: https://logstash-docs_960.docs-preview.app.elstc.co/diff

Versioning notes:
* 10.3.0 - 10.3.3. --> 7.7
* 10.4.0 - 10.4.2 --> 7.7
* 10.5.0 - 10.5.1 --> 7.8
* 10.6.0 - 10.6.2 --> 7.9
* 10.7.0 --> 7.10

ECS note:
* Starting with 10.6.1, these docs include links to ECS docs. ECS versioning differs from both plugin versioning and stack versioning.  For now, the ECS link still uses `current`.

### DIFF
--- a/docs/versioned-plugins/outputs/elasticsearch-v10.7.0.asciidoc
+++ b/docs/versioned-plugins/outputs/elasticsearch-v10.7.0.asciidoc
@@ -1,6 +1,7 @@
 :plugin: elasticsearch
 :type: output
 :no_codec:
+:branch: 7.10
 
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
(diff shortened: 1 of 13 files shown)


docs/versioned-plugins/outputs/elasticsearch-v10.7.0.asciidoc
```diff
@@ -948,11 +946,12 @@ See https://www.elastic.co/blog/elasticsearch-versioning-support.
...
 See https://www.elastic.co/blog/elasticsearch-versioning-support.
-See also https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-index_.html#_version_types
+See also https://www.elastic.co/guide/en/elasticsearch/reference/7.10/docs-index_.html#_version_types
 
 
 
 [id="{version}-plugins-{type}s-{plugin}-common-options"]
 include::{include_path}/{type}.asciidoc[]
 
+
```
karenzone, line 956:
```suggestion
```

yaauie (approved):
LGTM 👍

PR #964, opened 2020-11-05
Replace paths with attributes

karenzone:
Some links are hard coded to current, resulting in bad cross-document links.
This work updates the path to use the branch attribute to gracefully resolve links.

### DIFF
--- a/docs/plugins/filters/elasticsearch.asciidoc
+++ b/docs/plugins/filters/elasticsearch.asciidoc
@@ -171,7 +171,7 @@ Example:
 
 Authenticate using Elasticsearch API key. Note that this option also requires enabling the `ssl` option.
 
-Format is `id:api_key` where `id` and `api_key` are as returned by the Elasticsearch https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[Create API key API].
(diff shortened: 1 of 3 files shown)


docs/plugins/filters/elasticsearch.asciidoc
```diff
@@ -171,7 +171,7 @@ Example:
 
 Authenticate using Elasticsearch API key. Note that this option also requires enabling the `ssl` option.
 
-Format is `id:api_key` where `id` and `api_key` are as returned by the Elasticsearch https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html[Create API key API].
+Format is `id:api_key` where `id` and `api_key` are as returned by the Elasticsearch https://www.elastic.co/guide/en/elasticsearch/reference/{branch}/security-api-create-api-key.html[Create API key API].
```
dedemorton, line 174: I'm not sure why the paths are hard-coded here. Why not use the attributes in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc, which include the branch attributes? So  you would have `{ref}/security-api-create-api-key.html[Create API key API]`

dedemorton (approved):

PR #973, opened 2020-11-19
TEST VPR attributes and branchsets - DO NOT MERGE

karenzone:
We have mechanisms in place to handle plugin branch versioning in the Logstash Reference Guide. The shared attributes pick up the appropriate stack version (7.10, 7.11, 7.x, etc.)

The plugin version/branch mismatch is an ongoing problem in the Versioned Plugin Reference where there is no concept of stack versions.  We've worked around this issue by setting branch to `current`, but that causes pain at release time when `current` cuts over to a branch that no longer has the target links.

This PR is designed to help us test and vet various approaches to stack versioning, including full attributes like `{ref}` and `{logstash-ref]`, links using `{branch}` attribute in path, and hard-coded paths with branches.

**PREVIEW:** https://logstash-docs_973.docs-preview.app.elstc.co/guide/en/logstash-versioned-plugins/versioned_plugin_docs/output-elasticsearch-index.html

### DIFF
--- a/docs/versioned-plugins/outputs/elasticsearch-v10.6.1.asciidoc
+++ b/docs/versioned-plugins/outputs/elasticsearch-v10.6.1.asciidoc
@@ -19,6 +19,46 @@ END - GENERATED VARIABLES, DO NOT EDIT!
 
 include::{include_path}/plugin_header.asciidoc[]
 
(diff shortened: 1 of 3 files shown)


karenzone:
Testing complete. Closing

PR #974, opened 2020-11-20
VPR: Add/modify overrides for stack doc attributes

karenzone:
These overrides with a `{branch}` attribute give us more flexibility in managing version mismatches between plugin and stack. Thanks for consulting on the best approach, @dedemorton

### DIFF
--- a/docs/versioned-plugins/index.asciidoc
+++ b/docs/versioned-plugins/index.asciidoc
@@ -10,9 +10,6 @@
 
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
-// Override ref and logstash-ref settings imported from shared/attributes.asciidoc
-:ref: http://www.elastic.co/guide/en/elasticsearch/reference/{branch}
-:logstash-ref: http://www.elastic.co/guide/en/logstash/{branch}
 
 [[logstash-plugin-reference]]
 = Logstash Versioned Plugin Reference
--- a/docs/versioned-plugins/index.asciidoc
+++ b/docs/versioned-plugins/index.asciidoc
@@ -10,8 +10,9 @@
 
 include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
 
-// Override logstash-ref setting imported from shared/attributes.asciidoc
-:logstash-ref: http://www.elastic.co/guide/en/logstash/current
+// Override ref and logstash-ref settings imported from shared/attributes.asciidoc
+:ref: http://www.elastic.co/guide/en/elasticsearch/reference/{branch}
+:logstash-ref: http://www.elastic.co/guide/en/logstash/{branch}
 
 [[logstash-plugin-reference]]
 = Logstash Versioned Plugin Reference


dedemorton:
Do you need to set a default value for branch so that it resolves correctly for any topics you don't want to override?

karenzone:
Already there (Line 6). Unless I'm not understanding your comment.

karenzone:
After I updated the overrides to include the `{branch}` attribute instead of hard-coded `current`, I realized that they mirrored the book attributes in https://github.com/elastic/docs/blob/master/shared/attributes.asciidoc. Now the local overrides are no longer needed. We can use the shared attributes with the `{branch}` attribute set to `current` as you have already done in Line 6.  

This approach will make fixing broken `current` links at minor releases so much easier and straightforward.  Thanks for taking time to talk through various approaches with me. I learn something new every time.

dedemorton:
So your plan is to override `:branch:` in the file later on if you need to fix a broken link? That makes sense, but you'll need to make sure that you reset `:branch:` at the end of the file, or all the files that follow it will inherit the overridden setting.

Will you reset the branch to `:branch: current`? If so, you'll lose any benefits you get by inheriting the branch setting (which makes me think it might be better to set it explicitly in index, unless there's some way to access that inherited setting later on when you need to reset `branch` to the inherited setting.

One thing that's not clear to me is how `branch` gets set. You don't import the shared version file. How does the build know to use "current"? Maybe that happens automagically now? Did Nik set up some magic that I'm not aware of?

karenzone:
> So your plan is to override :branch: in the book later on if you need to fix a broken link? That makes sense, but you'll need to make sure that you reset :branch: at the end of the file, or all the files that follow it will inherit the overridden setting.

Yes. When I need to link a plugin doc file to a specific stack version, I will set `branch` to an appropriate stack version at the beginning of the file, and reset branch to `current` at the end of the file. 

> Will you reset the branch to :branch: current? 

Yes. Only when needed, I'll set the branch to a stack version at the beginning of the file, and reset it to `current` at the end. 

> If so, you'll lose any benefits you get by inheriting the branch setting (which makes me think it might be better to set it explicitly in index, unless there's some way to access that inherited setting later on when you need to reset branch).

I had considered letting the branch version "spill over" to consecutive files IF they the same stack version applies. I'm leaning toward setting and clearing in each file to make the behavior and expectation more obvious to anyone else who might have to touch the generated files.  

> ... (which makes me think it might be better to set it explicitly in index...

It is set in `index.asciidoc` at Line 6. This approach works well for 95%+ of situations. The override to fix links in specific files is the fly in the ointment. 

> One thing that's not clear to me is how branch gets set. You don't import the shared version file. How does the build know to use "current"? Maybe that happens automagically now? Did Nik set up some magic that I'm not aware of?

Branch is set in `index.asciidoc` at Line 6. The shared version file approach works only when the originating file is stack versioned.  Plugin-to-stack versioning is only a problem for the VPR. The Logstash Ref is stack-versioned, and resolves gracefully.

dedemorton:
>I'm leaning toward setting and clearing in each file to make the behavior and expectation more obvious to anyone else who might have to touch the generated files.

I think that's the best approach.

>Branch is set in index.asciidoc at Line 6.

Apparently I need new eye glasses.

dedemorton (approved):
LGTM!

karenzone:
@dedemorton  Thanks for working through this with me and for helping us arrive at the best solution.  The VPR is an amazing piece of work, and our customers really appreciate it. And I appreciate YOU! <3

PR #975, opened 2020-11-25
updated docs for 7.10

logstashmachine:


karenzone (changes requested):
Regressions are causing docs-CI to fail.

PR #976, opened 2020-12-15
updated docs for 7.x

logstashmachine:


karenzone (changes requested):
Regressions causing docs failures must be addressed before merge.

karenzone (approved):
Regressions addressed.

PR #980, opened 2021-01-20
updated docs for 7.10

logstashmachine:


karenzone (changes requested):
This PR introduces some regressions causing docs CI to break.

karenzone (approved):
Files containing regressions removed. Builds cleanly and LGTM

PR #986, opened 2021-02-18
Add default override setting for ecs_version

karenzone:
Add comments that explain the how's and why's of version management
for the Versioned Plugin Reference


gtback (approved):
LGTM!

PR #991, opened 2021-02-19
updated docs for 7.x

logstashmachine:


karenzone (changes requested):
Deprecations in input-log4j are causing broken links that must be resolved

karenzone (approved):
Builds cleanly after file containing regressions was removed. LGTM

PR #995, opened 2021-03-04
updated docs for 7.11

logstashmachine:

PR #999, opened 2021-03-08
Updated plugin docs for 7.11 with regressions removed

karenzone:
This work starts with generated plugin docs (#995), and removes files containing regressions


karenzone (commented):
Builds cleanly with regression removed

PR #1001, opened 2021-03-18
updated docs for 7.12

logstashmachine:

PR #1003, opened 2021-03-18
Updated plugin docs for 7.12 with regressions removed

karenzone:
Starts with generated plugin docs PR (#1001), and removes file containing regressions.


karenzone (commented):

PR #1004, opened 2021-03-18
 Updated plugin docs for 7.x with regressions removed

karenzone:
Starts with #1002, and removes file containing regressions.


karenzone (commented):

PR #1005, opened 2021-03-23
updated docs for 7.x

logstashmachine:
Generated docs to test the impact of a second file in the input-tcp `/docs/` directory. All appears in order and as expected. The second file is ignored and appears to have no impact on the docgen process.


karenzone (changes requested):
Regressions are causing build failures. Do not merge!

karenzone:
Replaced by #1007. Closed without merge.

PR #1006, opened 2021-03-23
updated docs for 7.12

logstashmachine:


karenzone:
The only change picked up is a regression.  Closing without merge.

PR #1007, opened 2021-03-23
  Updated plugin docs for 7.x with regressions removed

karenzone:
Starts with #1005, and removes file containing regressions.

This PR is still part of the docs test described in #1005: 
"Generated docs to test the impact of a second file in the input-tcp /docs/ directory. All appears in order and as expected. The second file is ignored and appears to have no impact on the docgen process."


karenzone (commented):
Builds cleanly with regressions removed. LGTM

karenzone (commented):
Builds cleanly with regressions removed. LGTM

PR #1008, opened 2021-03-24
updated docs for master

logstashmachine:


karenzone (changes requested):
Contains regressions causing docs build failures (input-log4j). Do not merge

karenzone:
NOTE TO SELF (@karenzone):  Good experiment, but DELETING the file containing regressions won't work, because the action deletes the whole file.  We still need that file, but without the regressions.  

Closing without merge.

PR #1018, opened 2021-04-20
updated docs for master

logstashmachine:

PR #1019, opened 2021-04-20
 Updated plugin docs for master with regressions removed

karenzone:
Starts with #1018 


karenzone (commented):
Builds cleanly and renders correctly with regressions removed. LGTM

PR #1011, opened 2021-03-30
Add Agent to list of input plugins

karenzone:
**PREVIEW:** https://logstash-docs_1011.docs-preview.app.elstc.co/guide/en/logstash/master/plugins-inputs-elastic_agent.html

Related: https://github.com/elastic/logstash/issues/12794

Add placeholder file
Add include statement and entry in list of input plugins
Add and verify target pages to avoid broken links

PR #1021, opened 2021-04-22
VPR: Add input-elastic_agent landing page

karenzone:
The input-elastic_agent page in the Logstash Reference needs to link to the corresponding page in the Versioned Plugin Reference.  (Related: https://github.com/elastic/logstash-docs/pull/1011). 

This PR creates a static landing page. Note that the versioned documents/link won't be generated programmatically at the this time because docs tooling relies on a version bump and published gem as a trigger. We can evaluate that enhancement for later if we want to. 

**PREVIEW:** https://logstash-docs_1021.docs-preview.app.elstc.co/guide/en/logstash-versioned-plugins/current/input-elastic_agent-index.html

### DIFF
--- a/docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
+++ b/docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
@@ -13,6 +13,5 @@ The `input-elastic_agent` plugin is the next generation of the
 <<input-beats-index,input-beats plugin>>, and is based on the
 https://github.com/logstash-plugins/logstash-input-beats[the same github repository].
 
-<<input-beats-index,Beats plugins>> versioned v6.1.3 and higher have an {type}-{plugin} counterpart.
 To see which plugin version you have installed, run 
 `bin/logstash-plugin list --verbose`.
--- a/docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
+++ b/docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
@@ -13,6 +13,6 @@ The `input-elastic_agent` plugin is the next generation of the
 <<input-beats-index,input-beats plugin>>, and is based on the
 https://github.com/logstash-plugins/logstash-input-beats[the same github repository].
 
-<<input-beats-index,Beats plugins>> versioned v.6.1.3 and higher have an {type}-{plugin} counterpart.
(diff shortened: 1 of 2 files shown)


karenzone:
@andsel @roaksoax @jsvd Comments are welcomed.

docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+
+== Versioned {plugin} {type} plugin
+[subs="attributes"]
+++++
+<titleabbrev>{plugin}</titleabbrev>
+++++
+
+The `input-elastic_agent` plugin is the next generation of the
+<<input-beats-index,input-beats plugin>>, and is based on the
+https://github.com/logstash-plugins/logstash-input-beats[beats repository].
```
jsvd, line 14: "beats repository" can be confused with elastic/beats, so I suggest:

```suggestion
https://github.com/logstash-plugins/logstash-input-beats[the same github repository].
```

docs/versioned-plugins/inputs/elastic_agent-index.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+
+<<input-beats-index,Beats plugins>> versioned v.6.1.3 and higher have an {type}-{plugin} counterpart.
```
jsvd, line 16:
```suggestion
<<input-beats-index,Beats plugins>> versioned v6.1.3 and higher have an {type}-{plugin} counterpart.
```

jsvd: @karenzone maybe I'm not understanding but what is the goal with this sentence?

karenzone: The input-elastic_agent didn't exist before v6.1.3.  We're linking to the input-beats versions page, but earlier versions aren't applicable. I'll work on rewording to make this clearer.

karenzone: I removed that sentence. If a user runs `run bin/logstash-plugin list --verbose`, no versions earlier than v6.1.3 will be returned anyway. If they're looking for the installed version, then all should be well.

jsvd (approved):
LGTM

PR #1029, opened 2021-05-18
VPR: Remove link to file that doesn't exist yet

karenzone:
We're adding a new output plugin to integration-workplace_search. The file we're linking to doesn't exist yet. 
Resolution: Hardcode the entry for now, and replace with updated link when file goes live.


karenzone (commented):
Clean CI. This should fix the docs build.

PR #1032, opened 2021-05-25
Manual docgen to create new integration files

karenzone:
Manual build to test https://github.com/elastic/logstash/pull/12925.

### DIFF
--- a/docs/plugins/filters/fingerprint.asciidoc
+++ b/docs/plugins/filters/fingerprint.asciidoc
@@ -5,9 +5,9 @@
 ///////////////////////////////////////////
 START - GENERATED VARIABLES, DO NOT EDIT!
 ///////////////////////////////////////////
-:version: v3.2.3
-:release_date: 2021-03-11
-:changelog_url: https://github.com/logstash-plugins/logstash-filter-fingerprint/blob/v3.2.3/CHANGELOG.md
+:version: v3.2.4
+:release_date: 2021-05-14
+:changelog_url: https://github.com/logstash-plugins/logstash-filter-fingerprint/blob/v3.2.4/CHANGELOG.md
 :include_path: ../../../../logstash/docs/include
 ///////////////////////////////////////////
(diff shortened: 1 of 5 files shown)


karenzone:
ToDo:  Research settings for the component plugins for the enterprise_search integration. The default_plugin setting for output-appsearch is coming from the metadata.json file. Figure out where the default setting for output-workplace_search is being picked up. 

Screen Shot 2021-05-25 at 7 19 11 PM

robbavey:
@karenzone - Did you build this from the latest changes in https://github.com/elastic/logstash/pull/12925? There was a time when both the integration plugin and the single app search output were in the metadata list (although I would have expected that to have given the opposite values to `default_plugin` as you see here)

The value for `elastic_workplace_search` should be `1` - the new integration plugin is set as `"default-plugins": true`, which should set that. But it should also set `app_search` to `1` too, so that is... interesting

karenzone:
I did build from the latest. The licensing change was in your last commit, and it's here in my terminal history:

```
Fast-forward
 rakelib/plugins-metadata.json                                                            | 8 ++++----
 tools/dependencies-report/src/main/resources/licenseMapping.csv                          | 1 +
 tools/dependencies-report/src/main/resources/notices/elastic-workplace-search-NOTICE.txt | 5 +++++
 3 files changed, 10 insertions(+), 4 deletions(-)
 create mode 100644 tools/dependencies-report/src/main/resources/notices/elastic-workplace-search-NOTICE.txt
```

robbavey (approved):
I'm ok with the manual change, but we should create an issue to track the weirdness around the incorrect setting of the default nature of the plugin

karenzone:
Thanks, robbavey.  I'll get these files in place so that I can hook them in. Then I'll generate the docs again to get us another data point on this default behavior.

karenzone:
> I'm ok with the manual change, but we should create an issue to track the weirdness around the incorrect setting of the default nature of the plugin

Here's the issue: https://github.com/elastic/logstash/issues/12939

PR #1044, opened 2021-06-03
Activate links

karenzone:
Add link to enterprise search page on elastic.co
Activate link from enterprise search integration to workplace search output


karenzone (commented):
Builds cleanly

PR #1045, opened 2021-06-03
Add file include statement and index entry for workplace search for 7.x

karenzone:


karenzone (commented):
Builds cleanly and gets the job done

PR #1048, opened 2021-06-04
Switch out basic plugin header with intg header for appsearch

karenzone:
Add link to enterprise search solution page to index page


karenzone (commented):
Builds cleanly and checkout out

PR #1052, opened 2021-06-04
Add entries for appsearch and workplace search

karenzone:
**PREVIEW:**  https://logstash-docs_1052.docs-preview.app.elstc.co/guide/en/logstash/master/output-plugins.html

The list of output plugins is quite long. While I was hooking up and testing the docs for the enterprise search integration, I kept looking for "workplace search" in the w's instead of the e's. Like... a lot.  I also found myself looking for app_search in the a's, but at least that's within scanning distance of "enterprise_app_search". 

This PR adds entries for both app_search and workplace_search so that a casual user doesn't overlook them. 

Another option would be to set up those entries like "See elastic_workplace_search," but why not get users to their intended destination without that extra click? And we're reinforcing the official "elastic" name in the description, so I think we're good.

### DIFF
--- a/docs/plugins/outputs.asciidoc
+++ b/docs/plugins/outputs.asciidoc
@@ -7,6 +7,7 @@ The following output plugins are available below. For a list of Elastic supporte
 
 |=======================================================================
 | Plugin | Description | Github repository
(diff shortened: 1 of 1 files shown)


roaksoax (approved):
LGTM!

PR #1060, opened 2021-06-14
This is a test to see how the PR checker behaves with a broken link

jsvd:


karenzone:
This looks like a good solution for catching doc problems in the VPR before passing them on to the elastic.co docs build, and is a big improvement over what we're doing now.  Auto-merge for passing PRs would be icing. 🍰 

https://elasticsearch-ci.elastic.co/job/elastic+logstash-docs+pull-request+build-docs/127/console

jsvd:
created https://github.com/elastic/docs-tools/pull/58 to no longer auto commit code but create PRs instead

PR #1064, opened 2021-06-21
Add Agent to list of input plugins for 7.x

karenzone:
Add placeholder file for input-elastic_agent file
Backports #1011 to 7.x


karenzone (commented):
Builds and renders as expected

PR #1065, opened 2021-06-21
updated docs for master

logstashmachine:

PR #1067, opened 2021-06-22
Update docs for master with regression removed

karenzone:
Starts with generated docs in #1065 
Removes input-app_search doc that would introduce regressions


karenzone (commented):
Builds cleanly and renders correctly with regression removed.

PR #1070, opened 2021-06-23
updated docs for master

logstashmachine:

PR #1072, opened 2021-06-23
Updated docs for master with regressions removed

karenzone:
Starts with generated docs (#1070)
Removes output-elastic_app_search doc that would introduce regressions (as noted in elastic/docs-tools#56 and elastic/docs-tools#57).


karenzone (commented):
Builds cleanly and renders as expected. LGTM.

PR #1108, opened 2021-07-16
Update agent with new variable and new v6.1.6 content

karenzone:
Manual update version info and content for input-elastic_agent at v6.1.6


karenzone (commented):

PR #1112, opened 2021-07-22
Remove link to limitations ea limitations

karenzone:
As product matures, limitations are being removed


karenzone (commented):

PR #1115, opened 2021-07-22
Remove link to ea limitations topic 7.x

karenzone:
As Elastic Agent matures, limitations are being removed. The topic targeted from the link will be going away, and so the link must be deleted, too.


karenzone (commented):

PR #1120, opened 2021-07-30
updated docs for master

logstashmachine:

PR #1124, opened 2021-08-03
Updated plugins for master with regressions removed

karenzone:
Starts with generated docs in #1120 


karenzone (commented):
Builds cleanly without regressions

PR #1125, opened 2021-08-03
Updated plugins for 7.x with regressions removed 

karenzone:
Starts with generated docs in #1121


karenzone (commented):
Builds cleanly and renders as expected with regressions removed

PR #1128, opened 2021-08-04
VPR: Add attribute for ecs default

karenzone:
## Release notes
[rn:skip] 

## What does this PR do?
Adds and links shared attribute file for accurately stating ecs default status for the Versioned Plugin Reference.

### Design intent/usage
We can add/update text related to ecs defaults (or other info we want to write once and use multiple places). Then, we can  reference it by calling the attribute. In this case `{ecs-default}`.

**Important:** This work mirrors a corresponding change for the Logstash Reference. 

## Potential gotchas
The Versioned Plugin Reference isn't stack versioned. This detail could make  consistent messaging across plugins and versions tricky.  If we apply this attribute and then change the text, the new text will be applied to _every_ version (past, present, and future) of every plugin using the attribute.  

One possibility:  introduce a new attribute for each major change to messaging, and replace the old attribute with the new one in the source file. Then the updated messaging will be applied only to applicable plugin versions and newer. 

### ToDo
- [ ]  Add note to attribute file describing how to manage [lack of] versioning for attributes in the VPR. (See previous section.) 

Related: https://github.com/elastic/logstash/pull/13083

**Test set:** https://github.com/elastic/logstash-docs/pull/1161
**PREVIEW:** https://logstash-docs_1161.docs-preview.app.elstc.co/guide/en/logstash-versioned-plugins/current/v1.3.0-plugins-inputs-azure_event_hubs.html#_test_attribute_here


docs/versioned-plugins/include/attributes-ls-vpr.asciidoc
```diff
@@ -0,0 +1,4 @@
+// These VPR settings correspond to the LSR settings in logstash/docs/include/attributes-ls.asciidoc
+// When we update one, we must update the other
+
+:ecs-default: Any value in {logstash-ref}/logstash-settings-file.html[`logstash.yml`] for pipeline.ecs_compatibility other than disabled is considered BETA and unsupported until Logstash 8.0 and the final 7.x version are released. Having this pipeline level flag set will cause even patch-level upgrades to automatically consume breaking changes in the upgraded plugins, changing the shape of data the plugin produces.
```
yaauie, line 4: Same as noted in upstream [PR comment](https://github.com/elastic/logstash/pull/13083#discussion_r717034110).

yaauie (approved):
🌮 LGTM

PR #1142, opened 2021-08-25
Update docs for 7.14 with regression removed

karenzone:
Starts with generated docs in #1140 
Removed input-beats doc that would introduce regressions and cause downstream issues when we switch the attributes for input-elastic_agent. 

TIP: Note that the `docs/plugins/inputs/beats.asciidoc` file in #1140 contains content updates, but no version bump. That's a red flag that this file might require further investigation.


karenzone (commented):

PR #1148, opened 2021-08-31
Update elastic_agent to v6.2.0 - 7.x

karenzone:
Update input-elastic_agent with latest content (v6.2.0) from input-beats


karenzone (commented):
Builds cleanly and renders as expected. LGTM

PR #1152, opened 2021-09-03
updated docs for 7.14

logstashmachine:


karenzone (changes requested):
Do not merge!

karenzone:
This update contains only a regression that we don't want. Closing without merge.

PR #1158, opened 2021-09-16
Updated docs for 7.14 with regression removed

karenzone:
Input-beats contains a link to a file that was removed. This content is reused in input-agent, and that's where the issues would arise.


karenzone (commented):

PR #1182, opened 2021-10-15
updated docs for 7.15

logstashmachine:


karenzone (commented):
Do not merge. Contains only a regression.

PR #1201, opened 2021-11-08
updated docs for 7.16

logstashmachine:


karenzone (approved):
Regression removed. LGTM

PR #1203, opened 2021-11-08
updated docs for main

logstashmachine:


karenzone (approved):
Builds cleanly and renders as expected. LGTM

PR #1207, opened 2021-11-11
updated docs for 7.16

logstashmachine:


karenzone (changes requested):
Regression only. DO NOT MERGE

karenzone:
Regression only Closing without merge

PR #1212, opened 2021-11-11
Manual process: Update input-elastic_agent to v6.2.2

karenzone:
Update input-elastic_agent with latest content (v6.2.2) from input-beats

Input-elastic_agent is an alias of input-beats. Updating docs for elastic_agent is a manual process, consisting of copying content from the generated input-beats output to the input-elastic_agent output in logstash-docs.

#### IMPORTANT
Copy only from the :version: section (Line 11) to the bottom of the file. The custom variables for input-elastic_agent are different from input-beats, and must be preserved.


karenzone (commented):
Verified output. Builds cleanly and renders as expected.

PR #1263, opened 2022-01-06
updated docs for 6.8

logstashmachine:

PR #1264, opened 2022-01-06
Updated docs for 6.8 with regressions removed

karenzone:
Starts with docs generated in #1263 and removes regressions in input-file and output-elasticsearch. 
Replaces #1263


karenzone (commented):
Builds cleanly and renders as expected with regressions removed

PR #1265, opened 2022-01-06
ManualProcess: Update elastic_agent to v6.2.5 for main

karenzone:
Update input-elastic_agent with latest content (v6.2.5)
Copies generated input-beats content from #1261 to aliased input-elastic_agent


karenzone (commented):

PR #1269, opened 2022-01-15
Change master to main in input edit_urls

karenzone:
The `Edit` links on each page in our documentation send the user to our source files in github. We value contributions from our community, and want to make contributing as easy as possible.

Renaming `master` branch to `main` in several repos (including _all_ plugin repos) means that `Edit` links for plugin docs don't resolve correctly.  This PR makes the fix for all input plugins. 

Related: https://github.com/elastic/logstash/issues/13619

**PREVIEW:** https://logstash-docs_1269.docs-preview.app.elstc.co/diff

PR #1276, opened 2022-01-16
Change master to main for input edit_urls 8.0

karenzone:
Effectively backports #1269 to 8.0 (with conflict resolution)


karenzone (commented):
Links checked and changes validated

PR #1277, opened 2022-01-16
Change master to main for input edit_urls 7.17 

karenzone:
Replace view links with edit links for java core plugins 
Backports #1276 to 7.17


karenzone (commented):
Verified

PR #1283, opened 2022-01-24
Change master to main for plugin edit_urls 6.8

karenzone:
The Edit links on each page in our documentation send the user to our source files in github. We value contributions from our community, and want to make contributing as easy as possible.

Renaming master branch to main in several repos (including all plugin repos) means that Edit links for plugin docs don't resolve correctly. This PR makes the fix for all plugins in the 6.8 branch.

Related: elastic/logstash#13619


karenzone (commented):
Builds cleanly and renders as expected. Updated edit_urls check out.

PR #1289, opened 2022-01-26
VPR: Add branch attr to prevent broken links at 8.0 cutover

karenzone:
Plugin versioning is independent of Logstash/stack versioning. This PR sets (and then resets) an attribute to define a stack version that is appropriate for each particular version of the elasticsearch output plugin. 

We updated the content source at v10.8.2 and removed the offending link. Therefore, the broken link isn't a problem beyond v10.8.1. 

**PREVIEW** https://logstash-docs_1289.docs-preview.app.elstc.co/diff


karenzone:
@elasticmachine run elasticsearch-ci/docs

karenzone (commented):
Builds cleanly. Links resolve to an appropriate stack version rather than relying on `current`.

PR #1336, opened 2022-03-31
Updated docs for main with regressions removed

karenzone:
Starts with docs generated in #1331 and removes files containing outdated content


karenzone (commented):

PR #1340, opened 2022-04-05
Add skip-existing option to LSR doc generator.

mashhurs:
Currently, although there is no plugin version release, there might be a doc content change. For this case, we usually revert back to original version to avoid doc breakings  ([#45](https://github.com/elastic/docs-tools/issues/45)).
Examples: 
- https://github.com/elastic/logstash-docs/pull/1338/
- https://github.com/elastic/logstash-docs/pull/1339

To overcome, we introduced `skip-existing` option for LSR doc tool with [this PR](https://github.com/elastic/docs-tools/pull/72) and now we need add `skip-existing` flag when calling the doc generator.

### DIFF
--- a/.github/workflows/reference_docs.yml
+++ b/.github/workflows/reference_docs.yml
@@ -59,7 +59,7 @@ jobs:
         working-directory: ./docs-tools
         run: |
           bundle install --path=vendor/bundle
-          bundle exec ruby plugindocs.rb --output-path ../logstash-docs ../logstash/plugins_version_docs.json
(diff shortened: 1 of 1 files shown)


roaksoax (commented):
In practice the change is fairly straightforward, but, what is the intended consequence of this change? I do not see anything in the PR that explains to me what this does and what is the impact?

mashhurs:
> In practice the change is fairly straightforward, but, what is the intended consequence of this change? I do not see anything in the PR that explains to me what this does and what is the impact?

Thanks for letting know @roaksoax. Added the description.

roaksoax (approved):
lgtm!

PR #1350, opened 2022-04-19

logstashmachine:


karenzone (commented):
Deprecation notices don't comply with our deprecation standard, but I'm going to let this go through because the information is technically accurate. 
Deprecation concerns addressed in a follow-up PR: https://github.com/logstash-plugins/logstash-input-http/pull/154

PR #1372, opened 2022-05-20
VPR: Add include file containing version attrs

karenzone:
Test to see if we can de-emphasize the version attributes a bit so they don't distract the user.  The version attributes are needed to populate the link URLs, so they don't need to be so visible. 

### To Do
Note to self:  If this approach works, add implementation notes in this PR for posterity.

### DIFF
--- a/docs/versioned-plugins/include/6.x/stack_version.asciidoc
+++ b/docs/versioned-plugins/include/6.x/stack_version.asciidoc
@@ -1,0 +1,18 @@
+ifeval::["{versioned_docs}"=="true"]
+:branch: %BRANCH%
+:ecs_version: %ECS_VERSION%
+endif::[]
+
+////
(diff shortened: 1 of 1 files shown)


docs/versioned-plugins/include/6.x/stack_version.asciidoc
```diff
@@ -0,0 +1,18 @@
+ifeval::["{versioned_docs}"=="true"]
+:branch: %BRANCH%
+:ecs_version: %ECS_VERSION%
+endif::[]
```
karenzone, line 4: Add to plugin header file?

karenzone: Another idea: Add a version include file for each stack/ecs version.  Use cond'l logic to inject attributes only for VPR.

PR #1390, opened 2022-07-13
VPR: Remove conditionals from plugin headers

karenzone:
The plugin headers (for both standalone and integration plugins) contain conditional processing to support both Versioned Plugin Reference (VPR) plugins and the Logstash Reference (LSR). No conditional processing is needed because the VPR header files live in the `logstash-docs` repo and the LSR header files live in the `logstash` repo.  Because of the separate locations, we can use purpose-built header files for each guide and remove the conditionals that may be confusing to people. 

This PR: 
* Removes conditional statements and content that applies only to the Logstash Reference plugins. 
* Removes conditional statements surrounding content for the Versioned Plugin Reference so that it is always present. 

Related: https://github.com/elastic/logstash/pull/14352


jsvd (approved):
LGTM

PR #1406, opened 2022-08-04
Add intg elements for missing aws codecs

karenzone:
Cloudtrail and cloudwatch codecs weren't updated by docgen.  This PR brings existing plugin docs inline while we research the root cause.

Manual changes: 
* adds :Integration: attribute
*  updates version, releasedate. and changelog links
* replaces `plugin_header` with `plugin_header-integration`


karenzone (commented):

PR #1418, opened 2022-08-25
Updated docs 8.4 with regressions removed

karenzone:
Starts with system generated content in #1417.
Removes regressions for AWS component plugins: cloudtrail-codec and cloudwatch-codec.  

Needed fix: https://github.com/elastic/docs-tools/issues/79
This problem will happen every time we generate docs for affected branches (currently main and 8.4), necessitating a manual fix on my part.


karenzone (commented):
Builds cleanly and won't cause problems (after regressions removed)

PR #1432, opened 2022-09-22
auto generated update of versioned plugin documentation (with manual link fix)

logstashmachine:


karenzone (changes requested):
DO NOT MERGE! Some fixes required.

docs/versioned-plugins/outputs/elasticsearch-v11.9.1.asciidoc
```diff
@@ -0,0 +1,1205 @@
...
+The Elasticsearch action to perform. Valid actions are:
+
+- `index`: indexes a document (an event from Logstash).
+- `delete`: deletes a document by id (An id is required for this action)
+- `create`: indexes a document, fails if a document by that id already exists in the index.
+- `update`: updates a document by id. Update has a special case where you can upsert -- update a
+  document if not already present. See the `doc_as_upsert` option. NOTE: This does not work and is not supported
+  in Elasticsearch 1.x. Please upgrade to ES 2.x or greater to use this feature with Logstash!
+- A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
+  would use the foo field for the action. If resolved action is not in [`index`, `delete`, `create`, `update`], the event will not be sent to {es}, and instead either will be sent to the pipeline's <<dead-letter-queues,dead letter queue (DLQ)>> if it is enabled, or will be logged and dropped.
```
karenzone, line 398:
```suggestion
  would use the foo field for the action. If resolved action is not in [`index`, `delete`, `create`, `update`], the event will not be sent to {es}, and instead either will be sent to the pipeline's {logstash-ref}/dead-letter-queues.html[dead letter queue] if it is enabled, or will be logged and dropped.
```

docs/versioned-plugins/outputs/elasticsearch-v11.9.1.asciidoc
```diff
@@ -395,7 +395,7 @@ The Elasticsearch action to perform. Valid actions are:
   document if not already present. See the `doc_as_upsert` option. NOTE: This does not work and is not supported
   in Elasticsearch 1.x. Please upgrade to ES 2.x or greater to use this feature with Logstash!
 - A sprintf style string to change the action based on the content of the event. The value `%{[foo]}`
-  would use the foo field for the action. If resolved action is not in [`index`, `delete`, `create`, `update`], the event will not be sent to {es}, and instead either will be sent to the pipeline's <<dead-letter-queues,dead letter queue (DLQ)>> if it is enabled, or will be logged and dropped.
+  would use the foo field for the action. If resolved action is not in [`index`, `delete`, `create`, `update`], the event will not be sent to {es}, and will instead either be sent to the pipeline's {logstash-ref}/dead-letter-queues.html[dead letter queue] if it is enabled, or will be logged and dropped.
```
karenzone, line 399: Explanation:  The `<<anchor-text>>` link format works only _within_ a book.  So if we were linking from a topic in the Logstash Reference (`logstash`) repo to a topic in the Logstash Reference, this format would have been perfectly fine.  Plugin docs are different. 

Remember that our docs tooling creates content for both the Logstash Reference (LSR) and the Versioned Plugin Reference (VPR) from the plugin doc source file.  The Versioned Plugin Reference (VPR) has no concept of the internal `<<dead-letter-queues>>` anchor, and that's why docs-ci throws a link check error in the PR. 

Instead, we have to use the link format we use between docs. But we can use the docs shared attributes (like `{logstash-ref}` in this case). This expanded format doesn't break the LSR, but allows the VPR to know where to find the linked content.

karenzone (approved):
Formatting error corrected. Docs build cleanly and pass docs-ci.

docs/versioned-plugins/outputs/elasticsearch-v11.9.1.asciidoc
yaauie: 🤦🏼 this one is totally on me. I had Logstash core docs pulled up and copy/pasted the "link" reference without considering that these are built independently.

karenzone: How fortuitous! Thank you for the opportunity to explain. ❤️ 😄 
Now this is documented in the `logstash-docs` repo in additional to our internal docs processes.

PR #1451, opened 2022-11-18

logstashmachine:


karenzone (changes requested):
DO NOT MERGE!

karenzone:
Note that this process is working as designed.  Think of docs-ci as our backup parachute. 🪂 
We're failing docs-CI for this PR instead of automerging this content which would result in breaking docs builds for **_all_** Elastic docs. 

Granted, it would be better if we had front-end validation at the plugin repo level. We have an [open issue](https://github.com/elastic/docs-tools/issues/42), but it's currently ON HOLD for reasons outlined in the issue.

karenzone (commented):
Closing to save notes/status and to allow for new docgen

PR #1533, opened 2023-09-20
Add a workflow to create an empty plugin placeholder in VPR.

mashhurs:
### Description
We need an automation to create an empty index page when we introduce new plugin, especially `integration` type is complex in docs. This PR create a workflow which executes https://github.com/elastic/docs-tools/pull/97 newly introduced placeholder create script.


jsvd (approved):
LGTM

PR #1537, opened 2023-09-25
Update the calling script name for generating placeholder.

mashhurs:
### Description
- The `create_plugin_placeholder.rb` script name was changed to `generate_plugin_placeholder.rb` but it wasn't updated in workflow. This PR updates the script name.
- It also intentionally changes the plugin type input to empty to encourage users to type plugin type.


jsvd (approved):
LGTM

PR #1545, opened 2023-10-02
LSR: Add files and structure

karenzone:

PR #1546, opened 2023-10-02
LSR: Add files and structure for new integration (#1545)

karenzone:
We need these files in 8.10 (`current`) for the VPR links to target.  (The VPR is not versioned, and therefore we always point to `current`.


karenzone:
46810

PR #1568, opened 2023-11-21
Update `silence_errors_in_log example`

BBQigniter:
The example is unfortunately deprecated and does not work anymore. See https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1159 

Also it would be nice to have a list of possible values somewhere - I could not find any

### DIFF
--- a/docs/plugins/outputs/elasticsearch.asciidoc
+++ b/docs/plugins/outputs/elasticsearch.asciidoc
@@ -987,12 +987,12 @@ if enabled, script is in charge of creating non-existent document (scripted upda
 
 Defines the list of Elasticsearch errors that you don't want to log.
 A useful example is when you want to skip all 409 errors
-which are `document_already_exists_exception`.
+which are `version_conflict_engine_exception`.
 
 [source,ruby]
     output {
       elasticsearch {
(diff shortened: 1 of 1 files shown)


BBQigniter:
Hi @karenzone, I signed that thing right after this message popped up, not sure what went wrong.

karenzone:
@BBQigniter, in that case, thank you for re-signing it.  :-)  
I'll take a look.

karenzone:
@BBQigniter, thank you for reading our docs and helping us make them better  All plugin doc changes are sourced out of the plugin repo in this case [logstash-output-elasticsearch](https://github.com/logstash-plugins/logstash-output-elasticsearch), get generated into doc files, and output into the `logstash-docs` repo.
 
The change to the example was included in https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1159/files#diff-cae5619b3d18ec99c5ccd0a9f6de0c6d3f53343c64692444551a7d29da6863e7, and the change has been propagated to the Logstash Reference.

Related: I opened https://github.com/logstash-plugins/logstash-output-elasticsearch/issues/1177 to preserve your request for more examples.

PR #1579, opened 2023-12-20
Include elastic_integration plugin.

mashhurs:
### Description
Include `elastic_integration` plugin manually for now.

### DIFF
--- a/docs/plugins/filters/elastic_integration.asciidoc
+++ b/docs/plugins/filters/elastic_integration.asciidoc
@@ -1,0 +1,528 @@
+:plugin: elastic_integration
+:type: filter
+:default_plugin: 1
+
+///////////////////////////////////////////
+START - GENERATED VARIABLES, DO NOT EDIT!
+///////////////////////////////////////////
+:version: v0.1.3
+:release_date: 2023-12-20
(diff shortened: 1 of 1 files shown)


docs/plugins/filters/elastic_integration.asciidoc
```diff
@@ -0,0 +1,528 @@
+:plugin: elastic_integration
+:type: filter
+:default_plugin: 1
```
karenzone, line 3: @mashhurs, I know that you generated this locally for `main`.  This plugin will probably be bundled by default in `main`, but isn't yet.  Seems like maybe we should set the `default_plugin` value to `0` until then?  Please LMKWYT.

mashhurs: Ah thanks for catching this up, it is due to my local and let me update it.

mashhurs, line 3:
```suggestion
:default_plugin: 0
```

karenzone: Thanks for confirming. Nice to be sure that we're on the same page.

karenzone (approved):
LGTM. Thanks!

karenzone:
50423

PR #1585, opened 2023-12-22
Refine description for filter-elastic_integration

karenzone:
Related: https://github.com/elastic/logstash/issues/15715


karenzone (commented):
Tested and renders as expected

PR #1639, opened 2024-04-30
Add placeholder for integration-snmp page to LSR

karenzone:
When we publish the new integration plugin, the snmp-integration page will be created. This PR provides a target that we can link to for early validation.


karenzone (commented):
Placeholder needed to test/merge snmp-integration

karenzone:
7595 (although won't actually build until we merge the index change)

PR #1640, opened 2024-04-30
Add snmp integration entry  to VPR index (Test)

karenzone:
Manually add entry to VPR integrations index
TEST: Will the page persist or get stomped on subsequent VPR runs


karenzone (commented):
Passes CI.

karenzone:
7644

PR #1677, opened 2024-06-12
Add snmp link targets to 8.14 branch

karenzone:
The SNMP integration goes GA at 8.15, and will not be backported to earlier versions. 

Because the Versioned Plugin Reference is not stack versioned, all links from the Logstash Versioned Plugin Reference default to `current`.  For now `current`= `8.14`. 

This PR adds target content to the placeholder docs we already have in place.  These changes are required to keep the VPR building correctly before we release 8.15.0 and `current` = `8.15.0`.

### DIFF
--- a/docs/plugins/integrations/snmp.asciidoc
+++ b/docs/plugins/integrations/snmp.asciidoc
@@ -131,6 +131,68 @@ input {
 }
 ----
 
-// ToDo: Any considerations that we should point out? 
+[id="plugins-{type}s-{plugin}-import-mibs"]
+==== Importing MIBs
+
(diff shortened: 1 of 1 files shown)


docs/plugins/integrations/snmp.asciidoc
```diff
@@ -131,6 +131,68 @@ input {
 }
 ----
 
-// ToDo: Any considerations that we should point out? 
+[id="plugins-{type}s-{plugin}-import-mibs"]
+==== Importing MIBs
+
```
karenzone, line 136: Manually-added content.

karenzone:
9674

PR #1727, opened 2024-09-13
Update versioned plugin documentation for codec values

karenzone:
Related: https://github.com/elastic/logstash/issues/16450


karenzone:
14186

PR #1775, opened 2024-12-26
TEST crossdoc links for output-elasticsearch

karenzone:
### TEST: DO NOT MERGE

Related:  https://github.com/elastic/docs-tools/issues/42

I've tested docs locally for https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1197. 
A local test can catch:
- some improper asciidoc formatting (if it's serious enough to prevent a successful build). 
   Example: Using the same anchor text in more than one place.
- missing `include` or referenced files or bad links (within the plugin doc or from the Logstash Reference)

After we get a successful build, we can PREVIEW the doc to spot other formatting errors--not bad enough to fail, but ugly. 
  Example: Tables or bulleted lists not rendering properly. 

A full doc build against all of elastic.co docs is the best way to be sure that we don't have any **_hidden cross-doc links_** coming from other docs. Example: Elasticsearch Reference linking to the elasticsearch output plugin.  Fortunately, CI can handle this test case, hence this PR. 

Obviously, if we have bad links coming from other docs, those links will need to be updated in the other doc(s). 

Note to self @karenzone:  Remind people that this ad hoc build strategy doesn't fully test docs generated with VPR builds.  That is, even with this testing, we could still have introduced potential failures.  

#### About this test

This PR is a result of manually copying the body of the plugin doc from https://github.com/logstash-plugins/logstash-output-elasticsearch/pull/1197, and pasting it into the existing generated output in `logstash-docs`.  Then I let CI do the testing against our full Elastic.co doc set.

### DIFF
--- a/docs/plugins/outputs/elasticsearch.asciidoc
+++ b/docs/plugins/outputs/elasticsearch.asciidoc
@@ -326,8 +326,10 @@ When a string value on an event contains one or more byte sequences that are not
 [id="plugins-{type}s-{plugin}-options"]
 ==== Elasticsearch Output Configuration Options
 
-This plugin supports the following configuration options plus the
-<<plugins-{type}s-{plugin}-common-options>> and the <<plugins-{type}s-{plugin}-deprecated-options>> described later.
+This plugin supports these configuration options plus the <<plugins-{type}s-{plugin}-common-options>> described later.
+
+NOTE: As of version 12.0.0 of this plugin, a number of previously deprecated SSL settings have been removed.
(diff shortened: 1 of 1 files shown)


karenzone (commented):
Validated: https://logstash-docs_bk_1775.docs-preview.app.elstc.co/guide/en/logstash/master/plugins-outputs-elasticsearch.html#plugins-outputs-elasticsearch-obsolete-options

karenzone:
Test complete. 
Closing without merge so that the usual docgen workflow can do its job.

PR #1797, opened 2025-01-29
TEST: Validate asciidoc for integration filter

karenzone:
#### TEST: DO NOT MERGE


karenzone (commented):
Some reworked needed, but builds without error. 
Related: https://github.com/elastic/logstash-filter-elastic_integration/pull/243/files

PR #1828, opened 2025-04-02
TEST esinput content against 8.x

karenzone:
## TEST - DO NOT MERGE!

Related: https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/205

#### What's happening? 
* This PR uses the existing Asciidoc CI to validate content for [ES input #205](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/205).
* This PR is based on the `8.x` branch because plugin content is still ADOC based. 
   This introduces a limitation because 8.x ≠ 9.0. 
* **Restating the obvious:**  Any content changes tested here will have to be manually added to the original PR. 

**PREVIEW:** https://logstash-docs_bk_1828.docs-preview.app.elstc.co/guide/en/logstash/8.x/plugins-inputs-elasticsearch.html

#### Commit history explained
* **Commit 1** - Original content copied from [source doc](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/205) starting just below the generated content values, and pasted into the generated doc in `logstash-docs` starting just below the generated content. 
* **Commit 2** - Review comments.

### DIFF
--- a/docs/plugins/inputs/elasticsearch.asciidoc
+++ b/docs/plugins/inputs/elasticsearch.asciidoc
@@ -107,8 +107,6 @@ Common causes are:
 [id="plugins-{type}s-{plugin}-cursor"]
 ==== Tracking a field's value across runs
 
-NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future
-
 .Technical Preview: Tracking a field's value
 ****
 The feature that allows tracking a field's value across runs is in _Technical Preview_.
@@ -124,7 +122,7 @@ Examples include:
 The Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
 When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin records the value of that field for the last document retrieved in a run into
 a file. 
-(The file location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).
+(The file location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>.)
 
 You can then inject this value in the query using the placeholder `:last_value`. 
 The value will be injected into the query before execution, and then updated after the query completes if new data was found.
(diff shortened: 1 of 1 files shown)


docs/plugins/inputs/elasticsearch.asciidoc
```diff
@@ -94,10 +94,140 @@ The plugin logs a warning when ECS is enabled and `target` isn't set.
...
+This feature works best when:
+
+. the query sorts by the tracking field;
+. the timestamp field is added by {es};
+. the field type has enough resolution so that two events are unlikely to have the same value.
+
+It is recommended to use a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
+If the tracking field is of this data type, an extra placeholder called `:present` can be used to inject the nano-second based value of "now-30s".
+This placeholder is useful as the right-hand side of a range filter, allowing the collection of
+new data but leaving partially-searcheable bulk request data to the next scheduled job.
```
karenzone, line 132:
```suggestion
[id="plugins-{type}s-{plugin}-cursor"]
==== Tracking a field's value across runs

NOTE: experimental:[] `tracking_field` and related settings are experimental and subject to change in the future


.Technical Preview: Tracking a field's value
****
The feature that allows tracking a field's value across runs is in _Technical Preview_.
Configuration options and implementation details are subject to change in minor releases without being preceded by deprecation warnings.
****

Some uses cases require tracking the value of a particular field between two jobs.
Examples include:

* avoiding the need to re-process the entire result set of a long query after an unplanned restart
* grabbing only new data from an index instead of processing the entire set on each job.

The Elasticsearch input plugin provides the <<plugins-{type}s-{plugin}-tracking_field>> and <<plugins-{type}s-{plugin}-tracking_field_seed>> options.
When <<plugins-{type}s-{plugin}-tracking_field>> is set, the plugin records the value of that field for the last document retrieved in a run into
a file. 
(The file location defaults to <<plugins-{type}s-{plugin}-last_run_metadata_path>>).

You can then inject this value in the query using the placeholder `:last_value`. 
The value will be injected into the query before execution, and then updated after the query completes if new data was found.

This feature works best when:

* the query sorts by the tracking field,
* the timestamp field is added by {es}, and
* the field type has enough resolution so that two events are unlikely to have the same value.

Consider using a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
If the tracking field is of this data type, you can use an extra placeholder called `:present` to inject the nano-second based value of "now-30s".
This placeholder is useful as the right-hand side of a range filter, allowing the collection of
new data but leaving partially-searchable bulk request data to the next scheduled job.
```

docs/plugins/inputs/elasticsearch.asciidoc
```diff
@@ -94,10 +94,150 @@ The plugin logs a warning when ECS is enabled and `target` isn't set.
...
+* the query sorts by the tracking field,
+* the timestamp field is added by {es}, and
+* the field type has enough resolution so that two events are unlikely to have the same value.
+
+Consider using a tracking field whose type is https://www.elastic.co/guide/en/elasticsearch/reference/current/date_nanos.html[date nanoseconds].
+If the tracking field is of this data type, you can use an extra placeholder called `:present` to inject the nano-second based value of "now-30s".
+This placeholder is useful as the right-hand side of a range filter, allowing the collection of
+new data but leaving partially-searchable bulk request data to the next scheduled job.
+
+id="plugins-{type}s-{plugin}-tracking-sample"]
```
karenzone, line 141:
```suggestion
[id="plugins-{type}s-{plugin}-tracking-sample"]
```

karenzone:
Validation complete. 
[Source PR](https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/205) merged.

PR #1839, opened 2025-04-14
Research: Plugin docgen without docbuild

karenzone:
What would happen if we skip the docbuild step in the GitHub action so that we can get output in a PR? 

Could we push the generated plugin output into a PR? Then the docs-ci on the PR would run and fail because the `include_path` variable would be over-written with the invalid path from plugin source. BUT, we'd have some output to work with!  We'd have a definitive list of changed plugins, and fixing the path using GitHub suggestions would be easy.  

**Micro goal:** Get a PR that contains doc output that I can adjust in order to avoid all of the manual steps required in https://github.com/elastic/logstash-docs-md/issues/16.

### DIFF
--- a/.github/workflows/reference_docs90.yml
+++ b/.github/workflows/reference_docs90.yml
@@ -63,11 +63,6 @@ jobs:
         run: |
           bundle install --path=vendor/bundle
           bundle exec ruby plugindocs.rb --skip-existing --output-path ../logstash-docs ../logstash/plugins_version_docs.json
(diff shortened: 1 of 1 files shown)


mashhurs (approved):
Let's try! 
lgtm~

karenzone:
Thank you so much for your help with this, @mashhurs! Now, instead of an error log with file names and lots of manual fixes to make, I get a PR that I can make quick fixes on.

PR #1850, opened 2025-04-25
VPR: Use non-standard header for intg filter

karenzone:
**Classic docs preview for PR validation:** https://logstash-docs_bk_1850.docs-preview.app.elstc.co/guide/en/logstash-versioned-plugins/current/filter-elastic_integration-index.html


karenzone:
Ugh.  I think these setting will be overwritten the next time these plugins are generated.

PR #1853, opened 2025-04-29
Use non-standard header 8.19

karenzone:
Related: https://github.com/elastic/logstash-filter-elastic_integration/issues/286


karenzone (commented):
Validated

PR #1871, opened 2025-06-17
GH: Rename github workflow for pre-9.0 docgen

karenzone:
Rename original docgen GH workflow to disambiguate from new docgen workflow for 9.0 and later.


mashhurs (approved):
LGTM~

PR #1872, opened 2025-06-18
Realign include path for 9.0

karenzone:
Related: https://github.com/elastic/logstash/issues/17695

Updates `:include_path`  from changes in #1861 

It's true that these docs aren't being published anywhere at the moment, but it's important to keep docs aligned for the next phase of docs migration: html -> MD in `logstash-docs-md`


karenzone (commented):
Validated

PR #2046, opened 2026-04-21
Fix plugin docs actions

donoghuc:
Use udpated bundler config option in favor of deprecated `--path`.

Fixes:
- https://github.com/elastic/logstash-docs/actions/runs/24744416558
- https://github.com/elastic/logstash-docs/actions/runs/24744429979

### DIFF
--- a/.github/workflows/reference_docs-pre90.yml
+++ b/.github/workflows/reference_docs-pre90.yml
@@ -61,7 +61,8 @@ jobs:
       - name: Generate plugin docs
         working-directory: ./docs-tools
         run: |
-          bundle install --path=vendor/bundle
+          bundle config set path 'vendor/bundle'
+          bundle install
           bundle exec ruby plugindocs.rb --skip-existing --output-path ../logstash-docs ../logstash/plugins_version_docs.json
       - name: Build docs
         working-directory: ./logstash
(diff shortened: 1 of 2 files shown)


.github/workflows/reference_docs-pre90.yml
```diff
@@ -61,7 +61,8 @@ jobs:
       - name: Generate plugin docs
         working-directory: ./docs-tools
         run: |
-          bundle install --path=vendor/bundle
+          bundle config set path 'vendor/bundle'
```
yaauie: Should we include the `--local` flag to disambiguate the behavior? I would prefer being explicit and not relying on an external tool making the right guess.

> Executing `bundle config set <name> <value>` defaults to setting `local` configuration if executing from within a local application, otherwise it will set `global` configuration.
> -- [Bundle Config: Set](https://bundler.io/man/bundle-config.1.html#set)

```suggestion
          bundle config set --local path 'vendor/bundle'
```

.github/workflows/reference_docs90.yml
```diff
@@ -61,7 +61,8 @@ jobs:
       - name: Generate plugin docs
         working-directory: ./docs-tools
         run: |
-          bundle install --path=vendor/bundle
+          bundle config set path 'vendor/bundle'
```
yaauie:
```suggestion
          bundle config set --local path 'vendor/bundle'
```

yaauie (approved):
Looks sensible. One suggestion to be more explicit.

